PR #64, opened 2025-11-13
update suggested spotify callback URL in readme

aritsune:
Spotify's dev site no longer allows http:// protocol callback urls unless their host is exactly 127.0.0.1 (for newly created apps, that is)


Nezreka:
Oh yes thank you for catching that.

PR #153, opened 2026-02-22
Update healthcheck URL in docker-compose.yml

chiefy:
Fix health check port in docker-compose


Nezreka:
Thanks for the PR, good catch!

PR #165, opened 2026-03-01
Fix spotify repeatable authorization calls

Regyl:
### Introduction
Hello dear SoulSync team. This is my first contribution to this project. I found your solution so interesting and would like to help this project growing.

### Problem
While run locally for the first time `web_server.py` I noticed server tried authorize Spotify with default `SpotifyClientID/SpotifyClientSecret` again and again. Looks like the root cause is poolling `/status` and other methods, which under the hood call `SpotifyClient.is_spotify_authenticated()`.

Also removed logs from git staging and other compile-time dependencies to improve contributing experience

### Solution
`SpotifyClient.is_spotify_authenticated()` now checks for cached auth token:

- If valid access/refresh token exists, return true
- Otherwise return false


Regyl:
@Nezreka please check when you will be free. And this one too please https://github.com/Nezreka/SoulSync/pull/168

PR #273, opened 2026-04-11
Use a multi-stage Dockerfile to reduce image size

kettui:
Build-time dependencies are omitted from the actual image, since they are not needed at runtime anymore.
This cuts the image size from 1.18GB to 759MB


kettui:
implemented via b3d3c01

PR #283, opened 2026-04-12
Setup logging early to prevent initial logs from being swallowed, fix db init operations

kettui:
Setup logging early to prevent initial logs from being swallowed
- There's a lot side-effects happening at import-time (eg. client initialization etc.), some of which include meaningful logging as well (eg. migration-related logging in MusicDatabase client).
- If we delay the logging initialization to the `__main__` block, we'll lose out on these early logs.

As a consequence of getting visibility on the early logs, I also found out that certain parts of the db init code would not run if the database was being created from scratch on a fresh setup. On the second startup, the logs would show this:
```
DEBUG - ID columns migration already applied
INFO - Added Deezer columns to discovery_pool table
INFO - Added similar_artist_deezer_id column to similar_artists table
INFO - Added album_deezer_id column to recent_releases table
INFO - Discovery tables created successfully
INFO - Added include_instrumentals column to watchlist_artists table
INFO - Added lookback_days column to watchlist_artists table
DEBUG - watchlist_artists table already has nullable spotify_artist_id or custom schema
INFO - Discogs enrichment columns added/verified
INFO - Added image_url, genres, popularity, metadata_updated_at columns to similar_artists for hero caching
INFO - Database initialized successfully
```

On the third startup things looked to be up-to-date finally.
The changes made in 1348de96c17f9b3253bb514e6cfea2aea8789863 make it so that the db initializes properly in one go.

This wasn't really relevant previously when the db init scripts were run continuously throughout the process' lifetime, but after [this](https://github.com/Nezreka/SoulSync/pull/276) change the init now runs only once at startup (as it should), so it can cause issues / confusion for new users if the initialization doesn't complete successfully in one go.


kettui:
(also removed the lingering, committed `music_library.db` file which was recently gitignored)

PR #295, opened 2026-04-14
Fix Socket.IO client fallback order

pxjx22:
## Summary
- prefer polling before websocket in the web UI Socket.IO client
- add a regression test to lock in the client transport order
- reduce the impact of failing websocket upgrades on the live UI

## Why
The current client forces websocket first. In the deployed app, websocket upgrade attempts can fail under the current Flask-SocketIO/Werkzeug runtime, which makes the UI feel stuck or constantly reconnecting even though the HTTP app stays healthy.

## Verification
- added a regression test for the transport order in 
- built and deployed the patched image privately on my instance
- verified the app stays healthy and the UI feels smoother with polling-first fallback

### DIFF
--- a/tests/test_websocket_client_transport.py
+++ b/tests/test_websocket_client_transport.py
@@ -1,0 +1,8 @@
+from pathlib import Path
+
+
+def test_websocket_client_prefers_polling_before_websocket():
+    script_path = Path(__file__).resolve().parents[1] / "webui" / "static" / "script.js"
+    script = script_path.read_text()
+
+    assert "transports: ['polling', 'websocket']" in script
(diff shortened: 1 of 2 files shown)


kettui:
I've noticed the feeling of  "getting stuck" as well. Curious if this was closed because it doesn't work as intended, or for some other reason?

Nezreka:
yeah I wanted to look into this more but hadn't gotten around to it. But it is something I do feel once in a while and do want to fix

pxjx22:
was closed be cause I have a fork with. whole bunch more bug fixes around the discovery pages and some non existent endpoint clean ups. happy to reopen. probably a rush close on my end

PR #298, opened 2026-04-14
Honor primary metadata source in album_completeness job and associated repair flow

kettui:
Album completeness and downstream repair flow now follow the configured primary provider first, with Discogs and Hydrabase support added alongside existing Spotify, iTunes, and Deezer paths.

I tried to keep the repetition and explicit if/else mapping to a minimum by introducing some `get_xxx_for_source` helpers, but I feel like ultimately some different abstraction layer should be added to unify these different operations under a single, shared interface that can be implemented by each provider easily, so that the different feature implementations wouldn't have to keep reinventing the wheel quite as much.

### DIFF
--- a/core/repair_jobs/album_completeness.py
+++ b/core/repair_jobs/album_completeness.py
@@ -256,7 +256,7 @@ class AlbumCompletenessJob(RepairJob):
             album_id = self._get_album_id_for_source(source, album_ids)
             if not album_id:
                 continue
-            api_tracks = self._get_album_tracks(context, source, album_id)
+            api_tracks = self._get_album_tracks(source, album_id)
             items = self._extract_track_items(api_tracks)
             if items:
                 return len(items)
@@ -288,7 +288,7 @@ class AlbumCompletenessJob(RepairJob):
             source_album_id = self._get_album_id_for_source(source, album_ids)
             if not source_album_id:
                 continue
-            api_tracks = self._get_album_tracks(context, source, source_album_id)
+            api_tracks = self._get_album_tracks(source, source_album_id)
             if self._extract_track_items(api_tracks):
                 break
 
@@ -353,14 +353,13 @@ class AlbumCompletenessJob(RepairJob):
     def _get_album_id_for_source(self, source: str, album_ids: dict) -> str:
         return album_ids.get(source, '')
 
-    def _get_album_tracks(self, context: JobContext, source: str, album_id: str):
+    def _get_album_tracks(self, source: str, album_id: str):
         """Fetch album tracks from a specific source."""
         try:
             if source not in ('spotify', 'itunes', 'deezer', 'discogs', 'hydrabase'):
                 return None
 
-            client = context.spotify_client if source == 'spotify' else None
-            return get_album_tracks_for_source(source, album_id, client=client)
+            return get_album_tracks_for_source(source, album_id)
         except Exception as e:
             logger.debug("Error getting %s album tracks for %s: %s", source.capitalize(), album_id, e)
             return None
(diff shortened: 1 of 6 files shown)


Nezreka:
Nice work on the column-alias refactor, that's a clear improvement. One thing to double-check: the current code explicitly guards against calling Spotify during a rate-limit ban (not context.is_spotify_rate_limited()) in both _get_expected_total and         
_find_missing_tracks. Does that check still exist somewhere in the new _get_album_tracks helper, or was it dropped? If it was
dropped, Spotify will get called during ban windows and the 429 will just get silently swallowed.

core/repair_worker.py
```diff
@@ -150,13 +154,12 @@ def set_metadata_enhancer(self, enhance_fn):
     # ------------------------------------------------------------------
     @property
     def spotify_client(self):
-        if self._spotify_client is None:
-            try:
-                from core.spotify_client import SpotifyClient
-                self._spotify_client = SpotifyClient()
```
kettui, line 156: The repair worker used a SpotifyClient instance of its own for seemingly no apparent benefit - by switching to use the shared `get_client_for_source` util, we'll both ensure that we're using an up-to-date client, as well as automatically ensuring that we're also a) authenticated, and b) not rate-limited, since the util checks for `is_spotify_authenticated` before returning the client

core/repair_jobs/album_completeness.py
```diff
@@ -331,6 +330,35 @@ def _get_settings(self, context: JobContext) -> dict:
...
+        """Return the active metadata source used for source prioritization."""
+        try:
+            return get_primary_source()
+        except Exception:
+            return 'deezer'
+
+    def _get_album_id_for_source(self, source: str, album_ids: dict) -> str:
+        return album_ids.get(source, '')
+
+    def _get_album_tracks(self, source: str, album_id: str):
```
kettui, line 343: > One thing to double-check: the current code explicitly guards against calling Spotify during a rate-limit ban (not context.is_spotify_rate_limited()) in both _get_expected_total and
_find_missing_tracks. Does that check still exist somewhere in the new _get_album_tracks helper, or was it dropped?

- `_get_expected_total` and `_find_missing_tracks` both defer to `_get_album_tracks` for their external service calls, which in turn calls the `get_album_tracks_for_source` helper to do the actual requests.
- Said helper then calls `get_client_for_source` to get the client for the requested source and does the search if a client was found.
  - If a Spotify request is attempted and we're rate-limited at the time, the `get_client_for_source` call will return None -> no request made

So effectively after my last few commits, explicit and repeated rate-limit checks should not be necessary in repair job implementations (or any other code for that matter), as long as we end up calling `get_client_for_source` shortly before attempting a Spotify request

core/repair_worker.py
kettui: It's worth noting that this `is_spotify_authenticated` look-up can go stale if we get rate-limited during the repair job execution, which is why most jobs are currently proactively re-checking the rate-limiting status in the first place.

If we stopped using this snapshot-based, context-driven (JobContext) spotify_client and switched to using `get_client_for_source` in the job workers instead, we'd stay up-to-date with the auth/rate-limiting status naturally and there might be less need for duplicated rate-limiting safeguards

core/repair_jobs/album_completeness.py
Nezreka: Great work overall, just one concern, the spotify_client property can now return None and may break other repair jobs that don't null-check it, im' pretty sure.

kettui: Earlier I checked that all repair jobs that reference the Spotify client from the context, have null checks included 👍

PR #301, opened 2026-04-15
Respect the primary metadata provider in more jobs

kettui:
## Summary

This branch removes the remaining Spotify-first assumptions from the repair jobs and aligns them with the configured metadata source order.

## What Changed

- **Unknown artist fixer** now resolves track IDs and title search using the shared metadata source priority, instead of always starting from Spotify.
- **Track number repair** now uses the same source priority and shared album-track lookup path, so Deezer and iTunes can be preferred over Spotify.
- **Missing cover art** now respects the configured primary source and explicit `prefer_source`, instead of defaulting to Spotify.
- **Metadata gap filler** now looks up ISRC data in a source-aware way, only fetches source track details when needed, and shows resolved source/track provenance in the UI.
- **Library reorganize** now uses the configured metadata source priority for album year lookups, and re-checks provider availability during the scan so Spotify can drop out cleanly if it becomes rate-limited.

## Notes

- This branch continues the earlier album completeness work by applying the same source-priority model to the remaining repair jobs.
- The goal is to keep repair behavior aligned with the user’s metadata provider settings and avoid unnecessary Spotify API usage.

### DIFF
--- a/core/repair_jobs/missing_cover_art.py
+++ b/core/repair_jobs/missing_cover_art.py
@@ -1,5 +1,6 @@
 """Missing Cover Art Filler Job — finds albums without artwork and locates art from APIs."""
 
+from core.metadata_service import get_client_for_source, get_primary_source, get_source_priority
 from core.repair_jobs import register_job
 from core.repair_jobs.base import JobContext, JobResult, RepairJob
 from utils.logging_config import get_logger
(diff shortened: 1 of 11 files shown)


core/repair_jobs/missing_cover_art.py
```diff
@@ -11,39 +12,68 @@
...
         'and apply it. The job does not download or embed artwork automatically.\n\n'
         'Settings:\n'
-        '- Prefer Source: Which API to try first for artwork (spotify or itunes)'
+        '- Prefer Source: Optional source to try first; otherwise the primary metadata source is used'
     )
     icon = 'repair-icon-coverart'
     default_enabled = True
     default_interval_hours = 48
-    default_settings = {
-        'prefer_source': 'spotify',
```
kettui, line 28: `prefer_source` is prioritized if specified, but otherwise we default to `get_primary_source()`

core/repair_jobs/missing_cover_art.py
```diff
@@ -88,24 +123,11 @@ def scan(self, context: JobContext) -> JobResult:
 
             artwork_url = None
 
-            # Check Spotify rate limit at the top level — skip Spotify entirely if banned
-            spotify_available = not context.is_spotify_rate_limited()
```
kettui, line 92: notice that we're not explicitly checking `is_spotify_rate_limited()` / `is_spotify_rate_limited()` anywhere in these jobs anymore, since that is implicitly checked when we call `get_client_for_source(source)`

PR #303, opened 2026-04-16
Consolidate watchlist scanning code, respect primary metadata provider

kettui:
## Summary

This branch makes web watchlist scanning respect the configured primary metadata source instead of defaulting to Spotify-first behavior.

## What Changed

- **Moved the shared watchlist scan loops into `core/watchlist_scanner.py`**
  - `web_server.py` now only handles triggers, progress/state, and post-scan orchestration.

- **Made watchlist discography lookup source-aware**
  - artist discography and image lookup now follow the configured metadata source priority
  - the scanner uses the first provider that can actually return albums

- **Made Spotify-only matching strict where needed**
  - Spotify ID backfill no longer falls through to fallback-provider results
  - Spotify cache hits still work even when Spotify is unavailable
  - fallback behavior stays available only where it is intentional

- **Preserved watchlist scan behavior**
  - placeholder tracklists are still skipped
  - Last.fm radio generation remains part of the post-scan flow
  - `get_artist_discography()` now routes through the shared helper correctly
  - removed dead watchlist scan entrypoints that were no longer used

## Notes

- This PR keeps the stricter source selection scoped to the watchlist scan path and the Spotify client behavior it depends on.
- More Spotify-heavy discovery helpers in `watchlist_scanner.py` are left for a follow-up PR.


core/spotify_client.py
```diff
@@ -1098,81 +1098,84 @@ def get_followed_artists(self) -> list:
...
+                    pass
+            if tracks:
+                return tracks
+
         use_spotify = self.is_spotify_authenticated()
 
         if use_spotify:
-            # Check Spotify cache
-            effective_limit = min(limit, 50)  # Spotify API max is 50
-            cached_results = cache.get_search_results('spotify', 'track', query, effective_limit)
```
kettui, line 1109: Spotify search cache was previously only read if spotify was enabled and not rate-limited. Now the cached results are prioritized even if rate-limits are up

core/spotify_client.py
```diff
@@ -1098,81 +1098,84 @@ def get_followed_artists(self) -> list:
             return []
 
     @rate_limited
-    def search_tracks(self, query: str, limit: int = 10) -> List[Track]:
-        """Search for tracks - falls back to configured metadata source if Spotify not authenticated"""
+    def search_tracks(self, query: str, limit: int = 10, allow_fallback: bool = True) -> List[Track]:
```
kettui, line 1101: added the option to disable the automatic fallback mechanism via `allow_fallback = False`, as in some cases it's highly important that we get results from Spotify only, and not some other provider (eg. when fetching spotify-specific information like id's etc.

PR #306, opened 2026-04-17
Refactor similar artist flows and incremental discovery pool updates in watchlist scanner

kettui:
## Summary
Continuation on recent PR's, tackling `_update_discovery_pool_incremental` and `_fetch_similar_artists_from_musicmap` next.

## What changed
- `_fetch_similar_artists_from_musicmap` now resolves similar artists through provider-priority searches and generic client helpers.
- Similar-artist backfill now updates fallback IDs for supported providers without the old iTunes-only helper path.
- `has_fresh_similar_artists` is now age-only, and the discovery-path comments/flow were aligned with the provider-priority model.
- Added and updated regression coverage for provider-priority matching, fallback backfill, and freshness behavior.

### DIFF
--- a/core/watchlist_scanner.py
+++ b/core/watchlist_scanner.py
@@ -30,63 +30,6 @@ DELAY_BETWEEN_ARTISTS = 4.0      # 4 seconds between different artists (was 2s, 
 DELAY_BETWEEN_ALBUMS = 0.5       # 500ms between albums for same artist
 DELAY_BETWEEN_API_BATCHES = 1.0  # 1 second between API batch operations
 
-# iTunes API retry configuration
-ITUNES_MAX_RETRIES = 3
-ITUNES_BASE_DELAY = 1.0  # Base delay in seconds for exponential backoff
-
-
-def _get_fallback_metadata_client():
-    """Get the configured metadata client — delegates to centralized metadata_service."""
-    from core.metadata_service import get_primary_source, get_primary_client
-    return get_primary_client(), get_primary_source()
-
-
-def itunes_api_call_with_retry(func, *args, max_retries=ITUNES_MAX_RETRIES, **kwargs):
-    """
-    Execute an iTunes API call with exponential backoff retry logic.
-
-    Args:
-        func: The function to call
-        *args: Arguments to pass to the function
-        max_retries: Maximum number of retry attempts
-        **kwargs: Keyword arguments to pass to the function
-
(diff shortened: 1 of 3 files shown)


database/music_database.py
```diff
@@ -7473,19 +7473,16 @@ def update_similar_artist_metadata_by_external_id(self, external_id: str, source
             logger.error(f"Error updating similar artist metadata by external ID: {e}")
             return False
 
-    def has_fresh_similar_artists(self, source_artist_id: str, days_threshold: int = 30, require_itunes: bool = True, require_spotify: bool = False, profile_id: int = 1) -> bool:
```
kettui, line 7476: `require_spotify` (and `require_itunes` for that matter) was removed because `has_fresh_similar_artists()` should answer only one question: “is this cache fresh enough to reuse?”

The old flag mixed freshness with provider coverage. That made the check behave like “fresh + enough Spotify IDs,” which forced refetches even when the cached similar artists were still recent and usable.

That behavior is now split:
- freshness is handled by `has_fresh_similar_artists()`
- fallback provider IDs are repaired by `_backfill_similar_artists_fallback_ids()`
- Spotify-specific consumers already have their own fallback paths when they need Spotify IDs

So this change removes an implicit Spotify bias from the freshness check and avoids unnecessary refetches while keeping the actual provider-resolution logic in the refresh/backfill paths.

core/watchlist_scanner.py
```diff
@@ -30,63 +30,6 @@
...
-ITUNES_BASE_DELAY = 1.0  # Base delay in seconds for exponential backoff
-
-
-def _get_fallback_metadata_client():
-    """Get the configured metadata client — delegates to centralized metadata_service."""
-    from core.metadata_service import get_primary_source, get_primary_client
-    return get_primary_client(), get_primary_source()
-
-
-def itunes_api_call_with_retry(func, *args, max_retries=ITUNES_MAX_RETRIES, **kwargs):
```
kettui, line 44: This legacy rate-limit retry helper was not carried over to the refactored implementation, as it was initially meant to be itunes-specific, while the current implementation is more generic / doesn't contain _that_ much provider-specific logic.

Retries driven by rate-limit errors would ideally be handled on a provider-basis at the client-level, so that a) provider-specific details can be taken into account appropriately, and b) features such as this don't have to keep reinventing the retry logic, and instead the clients can handle it in one place

PR #308, opened 2026-04-17
Fix generated IDs in repair flows & acoustid_scanner NPE

kettui:
## Summary

Fix repair paths that were creating database rows without explicit IDs after the `tracks.id` / `artists.id` schema moved to TEXT primary keys.

## What changed

- `incomplete_album` copy fills now generate a synthetic track ID before inserting the new row.
- Unknown-artist fixes now generate explicit artist IDs instead of relying on `lastrowid`.
- AcoustID scanning now skips legacy null-ID rows safely and sorts/checkpoints IDs defensively.
- Added a regression test covering the album-fill copy branch.

## Why

- Prevents new `NULL` IDs from being written into `tracks`.
- Avoids crashes in jobs that sort or update tracks by ID.
- Keeps repair jobs working consistently with the current string-ID schema.

### DIFF
--- a/core/repair_jobs/acoustid_scanner.py
+++ b/core/repair_jobs/acoustid_scanner.py
@@ -84,14 +84,16 @@ class AcoustIDScannerJob(RepairJob):
             checkpoint_id = context.config_manager.get(
                 f'repair.jobs.{self.job_id}.checkpoint_id', None
             )
+        if checkpoint_id is not None:
+            checkpoint_id = str(checkpoint_id)
 
         # Build ordered list of (track_id, info) sorted by ID for deterministic order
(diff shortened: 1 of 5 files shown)


kettui:
Noticed the null-pointer error after updating to latest and starting up the container, so decided to fix it.
The root cause for this turned out to be a bit wider (luckily not too wide), which was NULL track ids. Didn't realize null primary keys were a thing with sqlite, but TIL.

Patched the one feature that I found to be explicitly omitting the `id` field when inserting a new row. Existing rows should ideally be cleaned up, but I'll leave that for you to decide how to handle it @Nezreka. Likely a new migration of some sort.

PR #315, opened 2026-04-18
Run SoulSync under Gunicorn

kettui:
## Summary

Run SoulSync under Gunicorn instead of Flask’s built-in server, with a separate dev config for fast reloads and a cleaner Docker startup path.

## What changed

- Added a Gunicorn WSGI entrypoint in `wsgi.py`.
- Added production and development Gunicorn configs.
- Switched Docker and local startup to Gunicorn.
- Removed the direct `python web_server.py` server path.
- Added slow-request logging to help spot UI stalls.
- Tuned shutdown behavior so reloads and container exits stay responsive.
- Improved dev ergonomics so static assets refresh cleanly without restarting.

## Why

- Gunicorn is a better fit for a 24/7 Dockerized homelab app than Flask’s dev server.
- SoulSync uses in-process background work and Socket.IO, so a single-worker Gunicorn setup keeps behavior predictable.
- The new setup makes shutdown/reload behavior more stable without changing app behavior.

image

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -44,3 +44,5 @@ tidalapi>=0.7.6
 
 # WebSocket server for real-time UI updates
 flask-socketio>=5.3.0
+gunicorn>=25.3.0
+simple-websocket>=1.1.0
--- a/Dockerfile
+++ b/Dockerfile
@@ -90,4 +90,4 @@ ENV UMASK=022
 
 # Set entrypoint and default command
 ENTRYPOINT ["/entrypoint.sh"]
-CMD ["python", "web_server.py"]
+CMD ["gunicorn", "-c", "gunicorn.conf.py", "wsgi:application"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -81,8 +81,6 @@ HEALTHCHECK --interval=30s --timeout=10s --start-period=60s --retries=3 \
 # Set environment variables
 ENV PYTHONPATH=/app
 ENV PYTHONUNBUFFERED=1
-ENV FLASK_APP=web_server.py
-ENV FLASK_ENV=production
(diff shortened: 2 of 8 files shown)


requirements.txt
```diff
@@ -44,3 +44,5 @@ tidalapi>=0.7.6
 
 # WebSocket server for real-time UI updates
 flask-socketio>=5.3.0
+gunicorn>=25.3.0
+simple-websocket>=1.1.0
```
kettui, line 48: https://flask-socketio.readthedocs.io/en/latest/deployment.html#gunicorn-web-server

Nezreka:
This is new to me. You think this move is a good call?

Nezreka:
from what I understand this wouldn't work on windows as python web-server.py?

kettui:
I can add back support for direct `python web_server.py` usage for those that might be directly using that currently in non-dockerized deployments, but for docker, there's no reason _not_ to use something like gunicorn for running the application. A WSGI server like Gunicorn is meant for running production servers, while Flask's built-in server (the one SoulSync is currently using) is intended for development.
Since SoulSync is a homelab project mainly with very little users per instance, this is likely not a big problem, but it's still an industry standard way of running actual software, which makes the app look more professional as well

Milkshakes00:
> from what I understand this wouldn't work on windows as python web-server.py?

@Nezreka I can confirm this did break Windows support - It makes calls for fcntl, which obviously doesn't exist in Windows.

The back support looks to be working, but documentation could probably use an update. If someone goes to run Python standalone on Windows per the instructions (to leverage gunicorn), they're going to have a bad time. A notation to just use the old python web_server.py if Windows would probably be a good move.

Thanks for the fantastic service!

PR #316, opened 2026-04-18
Reduce UI stalling during enhanced search, add caching & small concurrency optimizations

kettui:
This changes the enhanced search flow from "_immediately fire N requests in parallel for all possible metadata providers_" into "_fetch results from the primary metadata provider first, and then make targeted (and slightly staggered) searches to other available providers_."

- This shift not only produces visible results faster, but it reduces UI stalls when multiple parallel requests don't necessarily compete with one another when it comes to painting the UI.
- Not only this, results are now also cached (up to 10 minutes currently) to make consequent duplicate searches faster.
- Searches were also limited so that short search terms (1-2 characters long) don't end up attempting a search, with the goal of avoiding accidental / redundant searches.
  - This optimization is up for debate - I'm not sure how common it is that someone would want to search with 2 letters. It's possible, so it might be useful to drop the threshold to 1
- Provider-specific artist / album / track searches are also done in parallel to speed up the query further

### DIFF
--- a/webui/static/script.js
+++ b/webui/static/script.js
@@ -9049,16 +9049,25 @@ function initializeSearchModeToggle() {
         const sources = data.sources || {};
         const primary = data.primary_source || 'spotify';
 
-        // Build tab list: primary first, then alternates sorted alphabetically
+        // Build tab list: primary first, then alternates sorted alphabetically.
+        // Hide completed zero-result sources so the bar stays focused.
         const sourceNames = Object.keys(sources).filter(s => sources[s].available);
-        if (sourceNames.length <= 1) {
+        const visibleSources = sourceNames.filter(name => {
+            const src = sources[name] || {};
+            const count = name === 'youtube_videos'
+                ? (src.videos?.length || 0)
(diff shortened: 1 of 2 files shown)


webui/static/script.js
```diff
@@ -18166,10 +18192,26 @@ function _gsRenderTabs() {
...
-    if (sources.length < 2) { el.style.display = 'none'; return; }
-    const labels = { spotify: 'Spotify', itunes: 'Apple Music', deezer: 'Deezer', discogs: 'Discogs', hydrabase: 'Hydrabase', youtube_videos: 'Music Videos' };
+    const labels = {
+        spotify: 'Spotify',
+        itunes: 'Apple Music',
+        deezer: 'Deezer',
+        discogs: 'Discogs',
+        hydrabase: 'Hydrabase',
+        youtube_videos: 'Music Videos',
+        musicbrainz: 'MusicBrainz',
```
kettui, line 18202: Noticed that musicbrainz was missing a label here

PR #318, opened 2026-04-18
Centralize metadata lookup for artist discography

kettui:
## Summary

This PR starts pulling artist discography resolution out of `web_server.py` and into `core/metadata_service.py` so metadata policy lives in one place instead of being reimplemented in the controller.

### What changed

- Introduced `MetadataLookupOptions` as a shared lookup policy object for metadata calls.
- Moved artist discography resolution into `core/metadata_service.get_artist_discography()`.
- Kept `web_server.py` as a thin controller layer that parses request args and returns JSON.
- Centralized source selection and enablement logic for Hydrabase in `metadata_service`.
- Added focused tests for:
  - discography source fallback behavior
  - artist name resolution behavior
  - source override behavior
  - Hydrabase enablement checks

## Why

The goal is to make metadata behavior easier to reason about, **test**, and extend. This PR lays the groundwork for moving more metadata-related logic out of `web_server.py` in future PRs. The driving motivation here again is to make the logic more testable, as testing the `web_server.py` file itself is not reasonable due to the amount of side-effects it currently has.

## Notes

- Discography lookup still uses the current best-effort name fallback behavior.
  - This behaviour has the potential to return the wrong discography, if an exact match is not found with the given artist name. Fixing this issue should be solved in a separate PR

### DIFF
--- a/web_server.py
+++ b/web_server.py
@@ -10941,154 +10941,105 @@ def get_artist_discography(artist_id):
     """Get an artist's complete discography (albums and singles)"""
     try:
         # Get optional artist name for fallback searches
-        artist_name = request.args.get('artist_name', '')
+        artist_name = request.args.get('artist_name', '').strip()
         # Optional source override from multi-source search tabs
-        source_override = request.args.get('source', '')
+        source_override = request.args.get('source', '').strip().lower()
 
         # Mirror to Hydrabase P2P network
         if hydrabase_worker and dev_mode_enabled and artist_name:
             hydrabase_worker.enqueue(artist_name, 'artist.albums')
 
-        # Determine which source to use
-        spotify_available = spotify_client and spotify_client.is_spotify_authenticated()
+        from core.metadata_service import (
+            get_artist_albums_for_source,
+            get_client_for_source,
+            get_primary_source,
+            get_source_priority,
+        )
 
-        # Import fallback client for non-Spotify lookups
-        fallback_client = _get_metadata_fallback_client()
-        fallback_source = _get_metadata_fallback_source()
+        def _search_artists_for_source(client, source: str, name: str, limit: int = 5):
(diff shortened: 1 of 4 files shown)


web_server.py
```diff
@@ -10941,247 +10937,79 @@ def get_artist_discography(artist_id):
...
-
-                if albums:
-                    print(f"Got {len(albums)} albums from {active_source} (source override)")
-            except Exception as e:
-                print(f"Source override ({source_override}) lookup failed: {e}")
-
-        # Try Hydrabase first when active (and no source override)
-        if not albums and _is_hydrabase_active() and artist_name:
-            try:
-                albums = hydrabase_client.search_discography(artist_name, limit=50)
```
kettui, line 11033: I chose to not carry over this hydrabase-specific `search_discography` "escape hatch", as it makes the code harder to reason about, imo. By omitting it, we can have more straight-forward code, since there doesn't have to be as many provider-specific exceptions. Instead we can go through the same flow for each provider, as in we can call the same `client.get_artist_albums` method for all clients consistently

web_server.py
```diff
@@ -10941,247 +10937,79 @@ def get_artist_discography(artist_id):
...
-        
-        print(f"Found {len(album_list)} albums and {len(singles_list)} singles for artist {artist_id}")
-        
-        # Debug: Log the final album list
-        for album in album_list:
-            print(f"Album: {album['name']} ({album['album_type']}) - {album['release_date']}")
-        for single in singles_list:
-            print(f"Single/EP: {single['name']} ({single['album_type']}) - {single['release_date']}")
-        
         # Gather artist enrichment info from cache + library
```
kettui, line 10977: artist enrichment info populating stays in the controller for now at least, but it probably makes sense to move it elsewhere as well later to make it testable

PR #328, opened 2026-04-19
Add URL-based deep linking for SPA navigation

### LINKED ISSUE #326: Feature Request: Deep Linking / URL-based Navigation
Problem: SoulSync is a single-page app where every page shares the same URL (/). This means:

You can't bookmark a specific page (Discover, Library, Artists, etc.)
Browser back/forward buttons don't navigate between pages
You can't share a direct link to a specific view
Refreshing the browser always lands you on the Dashboard instead of the page you were on
Proposed Solution: Add client-side routing so each page has its own URL path, e.g.:

/dashboard
/discover
/sync
/search
/artists
/automations
/downloads
/import
/library
/settings
/issues
This could be implemented with history.pushState/popstate (no framework needed) or hash-based routing (/#/discover). On page load, read the URL and navigate to the matching page. The server just needs a catch-all route that serves index.html for all paths.

Bonus: Sub-pages and state could also be encoded, like /artists/Linkin%20Park or /discover?section=recent-releases, and direct links to specific playlists, albums, or artist pages.

JohnBaumb:
Fixes #326

Every page in the app currently shares the URL /, so browser back/forward, bookmarks, link sharing, and refresh all lose navigation state. This PR adds URL-based routing on top of the existing SPA navigation without changing the page-switching logic itself.

Server:

New Flask catch-all route /path:page serves index.html so direct navigation to e.g. /discover works.
The catch-all calls abort(404) for paths beginning with api/, static/, auth/, callback, deezer/, tidal/, or status. Real routes still win via normal Flask dispatch; the abort only prevents unknown paths under those prefixes from falling through to index.html.
abort added to the existing flask import.

Client:

navigateToPage(pageId, options) now calls history.pushState so the URL reflects the current page. An internal skipPushState option is used when navigation is triggered by popstate or initial load to avoid duplicate history entries.
popstate listener registered in initializeNavigation so browser back/forward switches pages without a reload.
New _getPageFromPath() helper reads window.location.pathname and validates against a whitelist of known page IDs. artist-detail and playlist-explorer fall back to their parent pages (artists / library) since they need runtime context that isn't in the URL.
loadInitialData() reads the URL on startup, respects isPageAllowed(), and uses history.replaceState to set the initial entry. Query string and hash are preserved so flows like ?setup=1 still work.
Tests

test_spa_deep_linking.py- 39 tests covering:

All valid page IDs serve index.html.
Real API, auth, callback, deezer, tidal, and status routes are not shadowed.
Unknown paths under reserved prefixes return 404 (no index.html leak).
POST/PUT to SPA paths return 405.
Query strings don't affect routing.
Follows the existing test pattern of replicating routes in a minimal Flask app rather than importing [web_server.py](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/560a9dba96/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (which pulls in Spotify/Soulseek/Plex init).

Notes

- Unknown top-level paths like /foo-bar now return 200 index.html (standard SPA behavior) rather than 404. The client router falls back to the dashboard.
- Permission gating is unchanged. Deep links still pass through isPageAllowed() and redirect to the profile home page when disallowed.
- No new auth surface.
- Navigating to /artist-detail or /playlist-explorer via direct URL falls back to /artists or /library respectively. These pages require runtime context (which artist or playlist was selected) that isn't encoded in the URL yet, so there's no way to restore that state from a URL alone. A future enhancement could encode the artist/playlist identity in the URL to support direct deep links to specific artist detail pages.

### DIFF
--- a/tests/test_spa_deep_linking.py
+++ b/tests/test_spa_deep_linking.py
@@ -1,0 +1,205 @@
+"""Tests for the SPA deep-linking catch-all route.
+
+The real web_server.py cannot be imported at test time (it initializes
+Spotify, Soulseek, Plex, etc.), so we replicate the routing behavior in a
+minimal Flask app that matches the real implementation verbatim.
+"""
+
+import pytest
+from flask import Flask, abort
+
+
+# ---------------------------------------------------------------------------
+# App factory — mirrors the routes added in web_server.py
+# ---------------------------------------------------------------------------
+
+def _build_app():
+    app = Flask(__name__)
+    app.testing = True
+
+    @app.route('/')
+    def index():
+        return 'INDEX_HTML', 200
+
+    @app.route('/<path:page>')
+    def spa_catch_all(page):
+        if page.startswith(('api/', 'static/', 'auth/', 'callback', 'deezer/', 'tidal/', 'status')):
+            abort(404)
+        return 'INDEX_HTML', 200
+
+    # Stand-ins for real routes so we can verify the catch-all does not shadow them
+    @app.route('/api/ping')
+    def api_ping():
+        return {'ok': True}, 200
+
+    @app.route('/auth/spotify')
+    def auth_spotify():
+        return 'AUTH_SPOTIFY', 200
+
+    @app.route('/callback')
+    def oauth_callback():
+        return 'OAUTH_CALLBACK', 200
+
+    @app.route('/tidal/callback')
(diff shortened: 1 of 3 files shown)


JohnBaumb:
Tested on my local server with my entire library.

PR #339, opened 2026-04-20
Refactor artist album track lookup to use source priority

kettui:
## Overview

This branch moves the album-track lookup flow out of `web_server.py` and into `core/metadata_service.py`, so album track retrieval now follows the configured metadata provider priority instead of defaulting to Spotify-first behavior.

It also simplifies the UI-facing endpoint to a single album-based route and adds test coverage around the new service-layer logic.

## What changed

- Moved album track resolution and formatting into `core/metadata_service.py`
- Added source-priority album lookup helpers for:
  - exact provider lookup
  - database album ID resolution
  - name-based fallback search
- Normalized album track responses in the service layer so the controller no longer handles provider-specific formatting
- Switched the frontend to call the unified `/api/album/<album_id>/tracks` endpoint
- Renamed the old Spotify-specific route handler to avoid name collisions
- Added tests for:
  - primary source priority behavior
  - database album ID resolution
  - fallback album name search
  - normalized track formatting

## Why

The old route logic was doing too much work in the controller layer and still leaned heavily toward Spotify when it was available. This refactor makes the behavior consistent with the rest of the metadata service patterns and gives us a clean place to test it.

## Testing

- Added unit tests in `tests/test_metadata_service_album_tracks.py`
- Built a docker image locally and tested on my real SoulSync instance

## Notes

- This includes a route rename from the old artist-scoped path to the new album-scoped path.
- The frontend call sites were updated accordingly, and there are no known external consumers of the old API route.

### DIFF
--- a/core/metadata_service.py
+++ b/core/metadata_service.py
@@ -148,6 +148,24 @@ def get_album_tracks_for_source(source: str, album_id: str):
         return None
 
 
+def get_album_for_source(source: str, album_id: str):
+    """Get album metadata for an exact source.
+
+    Returns a provider-normalized album dict or None.
+    No fallback swaps.
+    """
+    client = get_client_for_source(source)
+    if not client or not hasattr(client, 'get_album'):
+        return None
+
+    try:
+        if source == 'spotify':
+            return client.get_album(album_id, allow_fallback=False)
+        return client.get_album(album_id)
+    except Exception:
+        return None
+
+
 def get_artist_albums_for_source(
     source: str,
     artist_id: str,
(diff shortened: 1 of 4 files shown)


core/metadata_service.py
```diff
@@ -641,6 +673,317 @@ def _extract_track_items(api_tracks: Any) -> List[Dict[str, Any]]:
...
+            album_columns = {row[1] for row in cursor.fetchall()}
+
+            source_chain = list(get_source_priority(preferred_source or get_primary_source()))
+            override = (preferred_source or '').strip().lower()
+            if override:
+                source_chain = [override] + [source for source in source_chain if source != override]
+
+            source_columns = {
+                'spotify': ('spotify_album_id',),
+                'deezer': ('deezer_id', 'deezer_album_id'),
```
kettui, line 798: unsure if `deezer_album_id` exists somewhere, but I kept it as a fallback just in case. `deezer_id` seems to be more commonly used in the database

web_server.py
```diff
@@ -11492,223 +11492,50 @@ def get_artist_discography(artist_id):
...
-                    continue
-
-    except Exception as e:
-        logger.debug("Error resolving DB album ID %s: %s", album_id, e)
-    return None
-
-
-@app.route('/api/artist/<artist_id>/album/<album_id>/tracks', methods=['GET'])
-def get_artist_album_tracks(artist_id, album_id):
+@app.route('/api/album/<album_id>/tracks', methods=['GET'])
```
kettui, line 11495: The `artist_id` from the path was not actually needed for the logic, so it was dropped from the request path for clarity.

PR #340, opened 2026-04-20
CI/CD pipeline improvements: dev nightlies, linting, caching, cleanup

JohnBaumb:
## Summary

Overhaul of the CI/CD pipeline to support a dev branch workflow, add code quality checks, and improve build performance.

### New: Dev nightly build workflow (dev-nightly.yml)
- Builds Docker images from the `dev` branch on every push and nightly at 4 AM UTC
- Pushes to GHCR (ghcr.io) instead of Docker Hub -- free for public repos, no extra secrets needed
- Two tags per build: rolling `:dev` tag and a pinned `2.33-dev.20260420.abc1234` tag (version + date + short SHA)
- Runs compile + pytest before building -- broken code never gets an image
- Scheduled runs skip automatically if no new commits in the last 24 hours (saves CI minutes)
- Uses the built-in GITHUB_TOKEN for auth -- zero setup required

### New: GHCR cleanup workflow (cleanup-dev-images.yml)
- Runs weekly on Sundays to prune old dev nightly images
- Keeps the last 10 dev builds plus all named tags (:dev, :latest, version numbers)
- Prevents GHCR storage from growing unbounded

### New: Ruff linting in build-and-test
- Added `ruff` to requirements-dev.txt
- New pyproject.toml with a lenient config that catches real bugs (undefined names, flake8-bugbear) without drowning the codebase in style warnings
- Errors show inline on PR diffs via `--output-format=github`
- Tests are relaxed (B and F rules ignored in tests/)

### Updated: Main branch Docker workflow (docker-publish.yml)
- Now auto-builds `:latest` on every push to main (previously manual-only)
- Manual dispatch still works for version-tagged releases with Discord announcement
- Discord announcement only fires on manual version-tagged releases, not every push

### Updated: Build-and-test scope
- Now uses `branches-ignore: [main, dev]` so it only runs on feature/fix branches
- main and dev have their own workflows with compile + test built in, avoiding duplicate CI runs

### Updated: Docker layer caching
- Both Docker workflows now use GitHub Actions cache (cache-from/cache-to type=gha) instead of no-cache
- Multi-arch builds (amd64 + arm64) will reuse layers and build significantly faster on subsequent runs

### Updated: Version string includes commit hash
- SOULSYNC_VERSION now displays as e.g. `2.33+abc1234` at runtime
- Resolves from SOULSYNC_COMMIT_SHA env var (Docker) or git (local dev)
- Backup restore version comparison uses base version only to avoid false mismatch warnings

## Setup required from the repo owner

1. Create a `dev` branch from main
2. Optionally make the GHCR package public (Settings -> Packages -> soulsync -> Change visibility)
3. That is it -- no new secrets needed

## Image pull examples

```
# Stable release (Docker Hub, same as today)
docker pull boulderbadgedad/soulsync:latest

# Dev nightly (GHCR, rolling)
docker pull ghcr.io/nezreka/soulsync:dev

# Pinned dev build
docker pull ghcr.io/nezreka/soulsync:2.33-dev.20260420.abc1234
```

## Open questions for contributors

Looking for feedback on any of the following:

- Should the ruff config be stricter? Currently only catches bugs (F, E9, W6, B) and ignores unused imports/variables.
- Any additional CI steps worth adding? (e.g. pip-audit for dependency security scanning, coverage reporting)
- Preferred nightly schedule? Currently 4 AM UTC -- can adjust
- Should dev nightlies also push to Docker Hub, or is GHCR-only fine for dev builds?
- Any other workflows or automation the project could benefit from?

### DIFF
--- a/.github/workflows/dev-nightly.yml
+++ b/.github/workflows/dev-nightly.yml
@@ -58,23 +58,6 @@ jobs:
           python -m compileall api core database services scripts web_server.py wsgi.py beatport_unified_scraper.py
           python -m pytest
 
-      - name: Generate version tag
-        if: steps.recent.outputs.skip != 'true'
-        id: version
-        run: |
-          # Read base version from web_server.py
-          BASE=$(python -c "
-          import re
-          with open('web_server.py') as f:
-              m = re.search(r'_SOULSYNC_BASE_VERSION\s*=\s*\"(.+?)\"', f.read())
-              print(m.group(1) if m else 'dev')
-          ")
-          SHORT_SHA=$(echo "${{ github.sha }}" | cut -c1-7)
-          DATE=$(date -u +%Y%m%d)
-          TAG="${BASE}-dev.${DATE}.${SHORT_SHA}"
-          echo "tag=$TAG" >> "$GITHUB_OUTPUT"
-          echo "Image tag: $TAG"
-
       - name: Set up QEMU
         if: steps.recent.outputs.skip != 'true'
         uses: docker/setup-qemu-action@v4
@@ -105,4 +88,5 @@ jobs:
             COMMIT_SHA=${{ github.sha }}
           tags: |
             ghcr.io/${{ github.repository_owner }}/soulsync:dev
-            ghcr.io/${{ github.repository_owner }}/soulsync:${{ steps.version.outputs.tag }}
+            ${{ github.event_name == 'schedule' && format('ghcr.io/{0}/soulsync:nightly', github.repository_owner) || '' }}
+            ${{ github.event_name == 'schedule' && format('ghcr.io/{0}/soulsync:nightly', github.repository_owner) || '' }}
--- a/.github/workflows/dev-nightly.yml
+++ b/.github/workflows/dev-nightly.yml
@@ -55,6 +55,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install -r requirements-dev.txt
+          python -m ruff check --output-format=github .
           python -m compileall api core database services scripts web_server.py wsgi.py beatport_unified_scraper.py
           python -m pytest
 
@@ -74,6 +75,13 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Generate build tag
+        if: steps.recent.outputs.skip != 'true'
+        id: tag
+        run: |
+          echo "date=$(date -u +%Y%m%d)" >> "$GITHUB_OUTPUT"
+          echo "short_sha=${GITHUB_SHA::7}" >> "$GITHUB_OUTPUT"
+
       - name: Build and push to GHCR
         if: steps.recent.outputs.skip != 'true'
         uses: docker/build-push-action@v7
@@ -88,5 +96,5 @@ jobs:
             COMMIT_SHA=${{ github.sha }}
           tags: |
             ghcr.io/${{ github.repository_owner }}/soulsync:dev
-            ${{ github.event_name == 'schedule' && format('ghcr.io/{0}/soulsync:nightly', github.repository_owner) || '' }}
+            ghcr.io/${{ github.repository_owner }}/soulsync:dev-${{ steps.tag.outputs.date }}-${{ steps.tag.outputs.short_sha }}
             ${{ github.event_name == 'schedule' && format('ghcr.io/{0}/soulsync:nightly', github.repository_owner) || '' }}
--- a/.github/workflows/dev-nightly.yml
+++ b/.github/workflows/dev-nightly.yml
@@ -1,0 +1,108 @@
+name: Dev Nightly Build
+
+on:
+  # Nightly at 4:00 AM UTC — only if dev branch has new commits
+  schedule:
+    - cron: '0 4 * * *'
+  # Also build on every push to dev for immediate feedback
+  push:
+    branches:
+      - dev
+  # Manual trigger for testing
+  workflow_dispatch:
+
+jobs:
+  nightly:
+    runs-on: ubuntu-latest
+    # Skip scheduled runs if dev branch has no new commits in the last 24h
+    # (pushes and manual triggers always run)
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout dev branch
+        uses: actions/checkout@v6
+        with:
+          ref: dev
+
+      - name: Check for recent commits (scheduled only)
+        if: github.event_name == 'schedule'
+        id: recent
+        run: |
+          LAST_COMMIT=$(git log -1 --format=%ct)
+          NOW=$(date +%s)
+          DIFF=$(( NOW - LAST_COMMIT ))
+          if [ "$DIFF" -gt 86400 ]; then
+            echo "No commits in the last 24h, skipping build"
+            echo "skip=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "skip=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Set up Python
+        if: steps.recent.outputs.skip != 'true'
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.11"
+          cache: pip
+          cache-dependency-path: requirements-dev.txt
+
+      - name: Install dependencies and run tests
+        if: steps.recent.outputs.skip != 'true'
+        env:
+          PYTHONPATH: ${{ github.workspace }}
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install -r requirements-dev.txt
+          python -m compileall api core database services scripts web_server.py wsgi.py beatport_unified_scraper.py
+          python -m pytest
+
+      - name: Generate version tag
+        if: steps.recent.outputs.skip != 'true'
+        id: version
+        run: |
+          # Read base version from web_server.py
+          BASE=$(python -c "
+          import re
+          with open('web_server.py') as f:
+              m = re.search(r'_SOULSYNC_BASE_VERSION\s*=\s*\"(.+?)\"', f.read())
+              print(m.group(1) if m else 'dev')
+          ")
+          SHORT_SHA=$(echo "${{ github.sha }}" | cut -c1-7)
+          DATE=$(date -u +%Y%m%d)
+          TAG="${BASE}-dev.${DATE}.${SHORT_SHA}"
+          echo "tag=$TAG" >> "$GITHUB_OUTPUT"
+          echo "Image tag: $TAG"
+
+      - name: Set up QEMU
+        if: steps.recent.outputs.skip != 'true'
+        uses: docker/setup-qemu-action@v4
+
+      - name: Set up Docker Buildx
+        if: steps.recent.outputs.skip != 'true'
+        uses: docker/setup-buildx-action@v4
+
+      - name: Log in to GHCR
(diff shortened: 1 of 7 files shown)


.github/workflows/dev-nightly.yml
```diff
@@ -0,0 +1,108 @@
+name: Dev Nightly Build
+
+on:
+  # Nightly at 4:00 AM UTC — only if dev branch has new commits
+  schedule:
+    - cron: '0 4 * * *'
```
kettui, line 6: I'm wondering about the usefulness of nightly releases - if releases are already made upon pushing to `dev`, does the nightly release provide any value on top of that? Wouldn't it essentially at best be a duplicate of whatever the latest version is?

kettui:
> Should the ruff config be stricter? Currently only catches bugs (F, E9, W6, B) and ignores unused imports/variables.

I personally like the current approach - start small, and work up the strictness gradually

> Any additional CI steps worth adding? (e.g. pip-audit for dependency security scanning, coverage reporting)

Dependency auditing does indeed sound like a good future addition. I think a prerequisite for that would be to make the dependency management more strict in general (ie. at the very least introducing a lockfile to make dependency versions consistent across builds). From what I'm seeing as I've been familiarizing myself with the Python scene through this project, `uv` seems like a popular pick in 2026 and is something I'd personally gravitate towards.


> Should dev nightlies also push to Docker Hub, or is GHCR-only fine for dev builds?

Ultimately I'd personally strive for consistency by not mixing and matching Docker Hub and ghcr. From past experiences, I've never had issues with ghcr, while docker hub has caused problems through their more aggressive rate limiting etc., so I'd gravitate towards ghcr if I had a choice to make. Curious to hear other peoples' experiences related to this.

Instead of publishing dev builds to docker hub in addition to ghcr, I feel it would be nice if it was the other way around - pushing `latest` to ghcr also and updating the docs / templates to use ghcr, while eventually phasing out Docker Hub altogether.

kettui (commented):
Overall I think that these additions and changes are great and very much needed to bring stability to the release process / versioning policy. This gives people the opportunity to use a pinned version that will remain stable, while also making it possible to easily roll back to previous versions in the case that things suddenly break after updates.

Certain practicalities still would need to be ironed out after making these changes, even though these changes by themselves are already a big improvement compared to the current situation. Some questions that come to mind:
1. **Should the versioning schema be adjusted to also include a patch version?**
This would make it possible to better communicate what people can expect from any given release, eg. if only the patch version is bumped, the release is only expected to include bug fixes and/or polishing to existing features, while minor versions could signal new features being added (or removed). Major version bumps would be there strictly to indicate a breaking change that would require explicit action from the user

2. To build on the first point, in order to make (officially versioned) release generation convenient and not feel like a chore / something you forget to do, further automations would be needed to be made to accommodate this. The exact details as to how this would be done are TBD at this point, but I feel like it would be nice that any time changes from the `dev` branch are flushed to `main`, it would be an automated process to figure out what the next release version should be.
A practical way to achieve this is by using [conventional commits](https://gist.github.com/qoomon/5dfcdf8eec66a051ecd85625518cfd13) to signal the nature of any given commit, as then the new version could be derived based on changes (commits) that have been added since the last version, without requiring anyone to make that decision manually.
This would require some coordination / enforcement to pull off, but it's a very common practice these days so I'm sure it ultimately wouldn't be an issue to implement if the choice was made, especially given project's current status where the amount of contributors is very limited.

.github/workflows/dev-nightly.yml
JohnBaumb: It would provide image comparisons for example, what the image performed like last night before changes without checkout/rebuilding, providing a fallback for users normally on dev. Or pinning a version to a machine while another feature is being implemented causing issues.

JohnBaumb:
> Instead of publishing dev builds to docker hub in addition to ghcr, I feel it would be nice if it was the other way around - pushing latest to ghcr also and updating the docs / templates to use ghcr, while eventually phasing out Docker Hub altogether.

Agreed on the GHCR consistency point. The only concern is that existing users are already pulling from Docker Hub. Switching the canonical image location could break their instances or leave them stuck on an old version. Main should keep pushing to Docker Hub during any transition, and the existing update-check system could surface a migration notice for Docker Hub users. Not something to act on now, but worth planning for later.

.github/workflows/dev-nightly.yml
kettui: I'm maybe not understanding this correctly, but with this current proposed workflow, what is the difference between a nightly scheduled release, compared to the latest release that was created as a result of pushing to `dev`? Since both get tagged with the same version format, as well as getting the generic `dev` tag associated with them, how does one separate the nightly releases from the rest? Wouldn't there need to be a separate `:nightly` tag for this to be useful in a way that (as I understand) you're describing?

kettui:
> Main should keep pushing to Docker Hub during any transition, and the existing update-check system could surface a migration notice for Docker Hub users. Not something to act on now, but worth planning for later

Exactly. That's why I figured that it would be beneficial to modify the primary publish workflow from the get-go to also push images to ghcr (in addition to Docker Hub), so that the transition could be slowly started straight away. Existing users could be nudged towards ghcr via documentation / README / compose template updates, and explicit runtime notifications like you mention could be later added for those that haven't made the switch after some period of time.

.github/workflows/dev-nightly.yml
JohnBaumb: Good callout lol, the nightly would just duplicate the latest push. I've updated the workflow; push builds only tag :dev, while the scheduled 4 AM build exclusively tags :nightly. That way :nightly is a stable daily checkpoint and :dev is bleeding edge. Also dropped the pinned date+hash tags since the commit SHA is already baked into the image and visible in the UI.

JohnBaumb:
Fixed, main pushes now go to both, that's a good idea.

.github/workflows/dev-nightly.yml
```diff
@@ -0,0 +1,92 @@
...
+          push: true
+          pull: true
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          build-args: |
+            COMMIT_SHA=${{ github.sha }}
+          tags: |
+            ghcr.io/${{ github.repository_owner }}/soulsync:dev
+            ${{ github.event_name == 'schedule' && format('ghcr.io/{0}/soulsync:nightly', github.repository_owner) || '' }}
+            ${{ github.event_name == 'schedule' && format('ghcr.io/{0}/soulsync:nightly', github.repository_owner) || '' }}
```
kettui, line 100: this seems to be duplicated now 👀

.github/workflows/dev-nightly.yml
kettui: > Also dropped the pinned date+hash tags since the commit SHA is already baked into the image and visible in the UI.

The secondary, more accurate tag is still useful for cases where someone might want to roll back to a previous exact version, in case the latest `:dev` or some other non-exact version happens to break things

JohnBaumb:
cleanup-dev-images.yml:

- Added nightly to the ignore-versions regex so :nightly survives weekly cleanup

dev-nightly.yml:

- Added ruff check linting step before compileall
- Added "Generate build tag" step producing date + short SHA
- Changed pinned tag from duplicated :nightly line to dev-YYYYMMDD-<sha> format (e.g. dev-20260421-319f8b6)

JohnBaumb:
I should also mention that these are failing on my fork actions and should be cleaned up in a separate PR before this is merged.

CI ran ruff on the current main branch and found 10 pre-existing lint errors (not introduced by this PR):

beatport_unified_scraper.py
- L247, L564, L1486, L1496, L1497: F541 f-string without any placeholders
- L460, L713, L889: B007 Loop control variable not used within loop body
- L728: B007 Loop control variable `level` not used within loop body

api/playlists.py
- L141: F541 f-string without any placeholders

These are all trivial fixes (remove `f` prefix or rename unused loop vars to `_`).

JohnBaumb:
Created a PR to fix those linting errors #344

JohnBaumb:
#344 was merged, should lint clean now.

kettui:
Seems like this one could use a rebase as well, but other than that it looks good to me.
Some things to note before this is merged:
- the `dev` branch should be synced with the latest changes from `main`
- if this does get merged, active development should happen in `dev`

Some other suggestions, feel free to comment on any of this:
- changes from `dev` should ideally be flushed to `main` only after a certain period of time when there are meaningful feature additions and/or fixes that are tested enough to consider stable (if anything truly is stable atm :joy:)
  - preferably these merges to `main` should always be accompanied with some form of a version change (github release with a changelog + docker image version bump), so no more overwriting `2.3` 50+ times :crossed_fingers: Non-generic versions should be immutable by nature, so that they stay consistent no matter who pulls the image and when. `latest`, `dev` and `nightly` should be the only moving targets.
  - all this helps `main` (and the `:latest` tag) stay as something that people can expect to not break whenever they pull an update, and most of all keeping actual pinned image versions as something that doesn't change if they are pulled again later on
  - `:dev` and `:nightly` are for those that want everything asap, even if things can break

JohnBaumb:
Fully agree on all of this. Rebased onto latest main.

On the workflow side, the pipeline already supports this model:

- Pushes to dev auto-build :dev and a pinned dev-YYYYMMDD-sha tag to GHCR
- The 4 AM cron additionally tags :nightly for a stable daily checkpoint
- Pushes to main auto-build :latest to both Docker Hub and GHCR
- Version tags (e.g. 2.35) are only created via manual workflow_dispatch and are immutable -- they're stamped once and never overwritten
- Weekly cleanup prunes old dev-* pinned tags while preserving :dev, :nightly, :latest, and version numbers
So the release flow would be: dev accumulates changes, gets tested via :dev/:nightly, then when ready a merge to main rebuilds :latest and a manual dispatch stamps the version tag. No more overwriting 2.3 fifty times.

The one thing not automated yet is generating a GitHub Release with a changelog on version tag creation. That could be added as a follow-up with something like [softprops/action-gh-release](https://github.com/softprops/action-gh-release), or done manually for now since releases should be deliberate anyway.

@Nezreka 

Merge this PR into main however you prefer (merge commit or squash). The push to main will trigger the docker-publish workflow, which will build and push :latest to both Docker Hub and GHCR. This first run will likely fail since the GHCR login needs the packages:write permission on the repo -- GitHub should prompt you to allow it, or you may need to ensure GITHUB_TOKEN has write access under Settings > Actions > General > Workflow permissions.

Sync dev to main

- After the merge, reset dev to match main so both branches start from the same baseline. Alternatively, you can delete the dev branch and recreate it from main.
-  PRs from this point forward should go through the dev branch first, no more pushing direct to main.

JohnBaumb:
Also for reference, the new image locations:

Docker Hub (stable releases only)

- boulderbadgedad/soulsync:latest -- auto-built on push to main
- boulderbadgedad/soulsync:<version> -- e.g. boulderbadgedad/soulsync:2.35, manual dispatch only

GHCR (all builds)

- ghcr.io/nezreka/soulsync:latest -- auto-built on push to main
- ghcr.io/nezreka/soulsync:<version> -- e.g. ghcr.io/nezreka/soulsync:2.35, manual dispatch only
- ghcr.io/nezreka/soulsync:dev -- rolling, every push to dev + scheduled + manual
- ghcr.io/nezreka/soulsync:dev-20260421-c91e651 -- pinned per build (date + short SHA)
- ghcr.io/nezreka/soulsync:nightly -- rolling, scheduled 4 AM UTC cron only

PR #343, opened 2026-04-21
Tidy up logging across the app

kettui:
## Summary
This branch cleans up logging across the app so runtime output is more consistent, less noisy, and easier to follow.

## What changed

- Replaced `print()` calls with structured logger calls across the runtime code
- Removed redundant severity text from log messages
- Tuned log levels in `web_server.py` so noisy trace output stays at `debug`
- Added `SOULSYNC_LOG_LEVEL` as a startup override for logging level
- Kept the UI log-level setting persisted and respected at startup
- Unified runtime log file locations under the configured app log directory
- Filtered Gunicorn access logs to drop static assets and Socket.IO noise
- Matched Gunicorn log formatting more closely to the app log style
- Kept AcoustID logging on its own file, separate from `app.log`
- Collapsed several old multi-line trace patterns into single structured log lines

## Why

- Ensures that all relevant logs end in a file as well (`print` only ends up in stdout)
- Makes the main app logs easier to scan
- Preserves useful operational detail without flooding the console
- Gives a simple env-based way to override log level on demand

## Verification

- Ran `py_compile` checks on the touched modules during the sweep
- Spot-checked the main logging paths and Gunicorn config changes

## Notes

- Beatport scraper logs remain opt-in only
- A few verbose debug traces still exist in `core/`, but the worst legacy spam patterns were cleaned up in this branch

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -11,6 +11,8 @@ from utils.logging_config import get_logger
 logger = get_logger("config")
 
 class ConfigManager:
+    _VALID_LOG_LEVELS = frozenset({"DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"})
+
     def __init__(self, config_path: str = "config/config.json"):
         # Determine strict absolute path to settings.py directory to help resolve config.json
         # This handles cases where CWD is different (e.g. running from /Users vs /Users/project)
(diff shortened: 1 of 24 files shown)


web_server.py
```diff
@@ -33,31 +33,31 @@
...
-# Dedicated source reuse logger — writes to logs/source_reuse.log
+# Dedicated source reuse logger — writes alongside app.log in the configured log directory
 import logging as _logging
 import logging.handlers as _logging_handlers
 source_reuse_logger = _logging.getLogger("source_reuse")
 source_reuse_logger.setLevel(_logging.DEBUG)
 if not source_reuse_logger.handlers:
-    os.makedirs("logs", exist_ok=True)
     _sr_handler = _logging_handlers.RotatingFileHandler(
-        "logs/source_reuse.log", encoding="utf-8", maxBytes=5*1024*1024, backupCount=2
```
kettui, line 49: if `app.log` location was adjusted to something else besides the default one, these other log files would still end up in the default logs/ directory. Now the other logs end up in the same place

config/settings.py
```diff
@@ -289,18 +297,70 @@ def _load_from_database(self) -> Optional[Dict[str, Any]]:
...
+        """Overlay env and persisted log level preferences onto the loaded config."""
+        env_level = self._load_env_log_level()
+        if env_level:
+            config_data.setdefault("logging", {})["level"] = env_level
+            logger.info(f"Using log level from SOULSYNC_LOG_LEVEL: {env_level}")
+            return config_data
+
+        stored_level = self._load_stored_log_level()
+        if stored_level:
+            config_data.setdefault("logging", {})["level"] = stored_level
```
kettui, line 360: If the log level was modified via the UI, the persisted log level would not be respected if the app was restarted. This changes that

PR #344, opened 2026-04-21
Fix existing ruff lint errors (F541, B007)

JohnBaumb:
Fixes the 10 pre-existing ruff lint errors that #340 will start catching once its linting step lands.

Changes across two files:

beatport_unified_scraper.py
- Removed 5 f-string prefixes on strings with no placeholders (F541)
- Renamed 4 unused loop variables to underscore-prefixed (B007): i -> _i, level -> _level

api/playlists.py
- Removed 1 f-string prefix on a string with no placeholders (F541)

No logic changes, just lint cleanup so #340 merges green.

### DIFF
--- a/api/playlists.py
+++ b/api/playlists.py
@@ -138,7 +138,7 @@ def register_routes(bp):
         try:
             # Forward to the internal sync endpoint
             import requests as http_requests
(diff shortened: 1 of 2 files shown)


Nezreka:
I think we may need a rebase if i'm understanding correctly

PR #357, opened 2026-04-22
Add SoulSync Discover sync tab with ListenBrainz, progress tracking, and Navidrome push

JohnBaumb:
Adds a full Discover Sync tab to the Sync page, rebased onto the post-script.js-split dev branch.

What changed:
- New Discover tab on the Sync page with playlist cards, source badges, and compact layout
- ListenBrainz playlist integration with deduplication and auto-update toggle persistence
- Sync progress tracking showing matched/total counts on cards and in toasts
- Navidrome playlist push on batch completion (both V1 and V2 paths)
- Active download state display with polling resume on page reload
- Stuck-download detection for downloading and catch-all states
- Serialized sync queue to prevent concurrent backend contention
- Clickable playlist modal to preview tracks before syncing
- Empty-state handling with guidance to visit Discover page first
- Server push status tracking (pending/pushing/success/failed/skipped) stored in sync_history
- Expanded playlist push to all sync types (mirrored, youtube, tidal, deezer, spotify, listenbrainz, beatport) not just discover
- Renamed _push_discover_playlist_to_server to _push_playlist_to_server
- New server_push_status column on sync_history table with auto-migration
- New update_sync_history_push_status() DB method
- Batch history API now includes server_push_status in response

Post-split JS changes mapped to: discover.js, sync-services.js, stats-automations.js, pages-extra.js

This is PR 1 of 3 from the discover-sync-no-services-fix branch split. PRs 2 and 3 (batch queue system, downloads page fixes) will follow after this merges.


JohnBaumb:
One thing I wanted to ask before this is merged is, on the regular discover page (/discover), when they push "Sync" on a playlist:

Should we push them to the /sync -> SoulSync Discover page, possibly scrolling them and highlighting the one they selected, and they can sync from there?

Or, alternatively, just fire the sync like it would if they had clicked on it in the /sync -> SoulSync Discover page?

The benefits for the first method is that they would be directed to the area where the syncs happen and are reported in, with the option of "Keep Updated".

The second method is more of a fire once, unaware that the first area exists or is a similar implementation.

JohnBaumb:
Also, I was thinking of adding something like the "Force Download" found in other places in the app, where it would ignore the quality requirements, since these playlists change, and some tracks are obscure...some users might just want the quantity of tracks over quality for these "dynamic" playlists.

JohnBaumb:
- Added force download toggle per playlist.

- Went with fire and forget but also a toast message to link to the sync page tab.

Nezreka:
Oop sorry I merged another pr causing this to have some conflicts.

JohnBaumb:
Fixed

Nezreka:
thanks for the rebase.

a couple small things:

- `_push_playlist_to_server` only does anything for navidrome. plex/jellyfin users hit the `'failed'` branch every sync with no reason shown. short-circuit to `'skipped'` when the active server isn't navidrome.
- `get_current_seasonal_playlist` (~44351) hardcodes `spotify_track_id`/`itunes_track_id`. deezer/discogs/hydrabase users silently miss tracks.
- `pollDiscoverBatchFromTab` (discover.js 9334) starts a setInterval with no cleanup path — if the user navigates away mid-sync or the batch stalls, it fires every 3s forever.

core design is solid.

JohnBaumb:
I don't have plex/jellyfin setup for music on my server, but I'll see what I can do to resolve this,

I'll fix the other two now.

Nezreka:
Oh no worries on that part then, I've got all three setup already.

kettui:
I'm not sure if it would ultimately end up affecting things badly, but the branch still has a whole bunch of duplicated commits from Boulder that already exist in `dev`. Would likely be the cleanest choice to sync the upstream dev branch locally first, then rebase this branch on top of it while dropping the commits from Boulder in the process. Another easy solution would naturally be to squash the commits before merging, but I'm personally not a big fan of that as individual commits and their descriptions can hold a lot of useful information when looking back at some changes in the future

JohnBaumb:
rebased, much cleaner.

webui/static/style.css
```diff
@@ -59937,4 +59939,324 @@ body[data-artist-source="source"] #artist-detail-page #library-artist-enhance-bt
 
 #artist-detail-page .release-card.album-card .mb-card-icon:hover {
```
kettui, line 59940: Some AI review findings:

L59940-L59978: 🔴 bug: this block is malformed, so .discover-sync-card loses padding/display/gap/transition and the parser swallows the rest of the section. Move the stray declarations back inside the rule and close .mb-card-icon:hover before the Discover comment.

core/seasonal_discovery.py
```diff
@@ -371,8 +371,10 @@ def _search_discovery_pool_seasonal(self, season_key: str, source: str = 'spotif
             config = SEASONAL_CONFIG[season_key]
             keywords = config['keywords']
 
-            # Use the right track ID column based on source
-            track_id_col = 'spotify_track_id' if source == 'spotify' else 'itunes_track_id'
+            # itunes stores IDs in itunes_track_id; all other sources
+            # (spotify, deezer, discogs, hydrabase, etc.) use spotify_track_id
+            # as the generic ID column.
+            track_id_col = 'itunes_track_id' if source == 'itunes' else 'spotify_track_id'
```
kettui: L374-L377: 🔴 bug: every non-itunes source now reads spotify_track_id, but Deezer rows are stored in deezer_track_id and other non-Spotify sources still use itunes_track_id. Seasonal curation will miss those playlists entirely. Branch on the real source column.

web_server.py
```diff
@@ -43864,6 +44061,269 @@ def refresh_discover_data():
...
+            {'type': 'hidden_gems', 'name': 'Hidden Gems', 'description': 'Underappreciated gems from your pool', 'icon': '💎'},
+            {'type': 'discovery_shuffle', 'name': 'Discovery Shuffle', 'description': 'Random tracks from discovery', 'icon': '🔀'},
+            {'type': 'familiar_favorites', 'name': 'Familiar Favorites', 'description': 'Familiar tracks you love', 'icon': '❤️'},
+        ]
+
+        # Check if discovery pool has any data (needed for personalized playlists)
+        try:
+            with database._get_connection() as conn:
+                pool_count = conn.execute(
+                    "SELECT COUNT(*) FROM discovery_pool WHERE source = ?", (active_source,)
```
kettui: L44187-L44189: 🟡 risk: this count ignores profile_id, so another profile's discovery pool can make the current profile look populated. Add AND profile_id = ? here.

webui/static/stats-automations.js
```diff
@@ -2157,6 +2157,7 @@ function importFileSubmit() {
 // ── Mirrored Playlists ────────────────────────────────────────────────
 
 let mirroredPlaylistsLoaded = false;
+let discoverSyncPlaylistsLoaded = false;
```
kettui, line 2160: L2160: 🟡 risk: discoverSyncPlaylistsLoaded never resets, but init.js now clears discover pollers on page leave. After one visit, reopening Sync won't refetch Discover status and the tab can stay stale. Reset the flag on exit or reload on each discover-tab activation.

webui/static/style.css
```diff
@@ -59937,4 +59939,323 @@ body[data-artist-source="source"] #artist-detail-page #library-artist-enhance-bt
...
     opacity: 1;
+
```
kettui, line 59943: Still some issues - the artist-detail-page block above doesn't cut off here

webui/static/style.css
```diff
@@ -59937,4 +59939,323 @@ body[data-artist-source="source"] #artist-detail-page #library-artist-enhance-bt
...
+    font-weight: 400;
+    line-height: 1.4;
+}
+
+.discover-sync-card {
+    background: rgba(255, 255, 255, 0.04);
+    border: 1px solid rgba(255, 255, 255, 0.08);
+    border-radius: 10px;
+    margin: 3px 6px;
+}
```
kettui: Either an unexpected termination here, or there is some other definition missing below

webui/static/discover.js
```diff
@@ -8903,3 +8954,592 @@ if (document.readyState === 'loading') {
...
+    }, 3000);
+
+    // Register so page-leave cleanup can clear it
+    discoverSyncPollers[playlistType] = pollInterval;
+}
+
+/**
+ * Map a discover playlist type to its API endpoint for fetching tracks.
+ */
+function _discoverPlaylistApiUrl(playlistType) {
```
kettui, line 9471: > bug: _discoverPlaylistApiUrl() still has no build_playlist entry, so the existing Build Playlist Sync button (line 3428) now falls into Unknown playlist type. Add the build_playlist mapping back.

web_server.py
```diff
@@ -28923,8 +28923,22 @@ def _on_download_completed(batch_id, task_id, success=True):
                     except Exception:
                         pass
 
-                # Update YouTube playlist phase to 'download_complete' if this is a YouTube playlist
+                # Push playlists to media server after downloads complete
```
kettui: web_server.py (line 28926): 🔴 bug: _on_download_completed() runs once per task, so this starts _push_playlist_to_server() before the batch is done and can fire it multiple times. Keep the push only in _check_batch_completion_v2() or gate it on the final batch transition.

PR #361, opened 2026-04-23
Feature/unify search and artist detail

Nezreka:
## Summary

Consolidates the overlapping Search / Artists / Artist-Detail surfaces into a single unified flow.

- **Search page** — Enhanced/Basic toggle replaced with a single source-picker dropdown; Soulseek is now selectable alongside Spotify/iTunes/Deezer/Discogs/Hydrabase/MusicBrainz. Page id renamed `downloads` → `search`. Embedded Download Manager removed.
- **Inline Artists page retired** — `artists.js` deleted (~4K lines), `#artists-page` HTML removed, `/artists` aliases to `/search`. Rendering functions relocated to `library.js`; shared helpers extracted to `shared-helpers.js`.
- **Standalone `/artist-detail` is now the single destination** for both library and source artists. Big-photo album/single cards restored (matching the retired inline design). Backend enriches source artists with Last.fm bio/stats, source-side genres/followers, and per-service IDs for badge rendering. Discography variant-dedup disabled on source-only view so every release surfaces.
- **Library upgrade detection** — clicking a Deezer/Spotify/iTunes result for an artist already in the library routes to the library view (with owned releases, enrichment, completion bars) instead of a bare source view.
- **Smart back button** — labels and navigation target adapt to origin page ("Back to Search", "Back to Discover", etc.).
- **Streaming fallback** — Popular on Last.fm play buttons fall back to Soulseek/YouTube via `/api/enhanced-search/stream-track` when the track isn't in the local library.
- **Similar Artists** added to standalone /artist-detail with page-aware click navigation and lazy-loaded backgrounds.

### Fixes shipped alongside

- `no such column: deezer_artist_id` error on every watchlist-config GET — artists table uses `deezer_id`/`discogs_id`, watchlist_artists uses the `_artist_id` suffix; the query was mixing them.
- Streaming completion stuck on "Checking…" — handler still targeted old `.completion-text` markup.
- Date-dependent CI test on `test_update_discovery_pool_incremental_uses_source_priority` — hardcoded release dates replaced with a relative constant so the 7-day cutoff doesn't expire.

### Regression tests added

- `tests/test_artist_source_lookup.py` — schema-vs-query contract tests for `_SOURCE_ID_FIELD`, `_find_library_artist_for_source`, the watchlist-config enrichment query, and `_build_source_only_artist_detail` response shape.
- `test_metadata_service_discography.py` — new test for `MetadataLookupOptions.dedup_variants=False` branch.
- `test_script_split_integrity.py` — updated for `artists.js` removal and `shared-helpers.js` addition.

## Why

Three overlapping surfaces (Search / Artists sidebar / Artist Detail) were doing nearly the same job with subtly different UX — flagged as a new-user pain point in Discord community feedback (Cin). The Artists page and Search page both let you type an artist name, both showed external metadata results, both had artist-detail navigation. The standalone artist-detail page rendered a different design from the inline Artists page and 404'd on most actions for non-library artists. Net result: users learned three mental models, and clicking the same thing in different places led to different outcomes.

This PR collapses that to one Search entry point, one artist-detail destination, and one rendering style — while preserving the library-vs-source distinction where it actually matters (owned releases, enhanced editing, enrichment coverage).

## Test plan

- [ ] `python -m pytest` — full suite green
- [ ] **Library artist** → standalone page shows hero, badges, bio, genres, stats, albums, singles, Similar Artists, completion bars, Top Tracks, Enhance Quality button
- [ ] **Source artist** (e.g. search Deezer for Kendrick Lamar, click result) → standalone page shows hero with Last.fm bio, source-side genres/followers, correct service badge, discography with every release, Similar Artists — no Enhanced toggle, no library-only UI
- [ ] **Library upgrade** — click a Deezer result for an artist already indexed in Plex/Jellyfin → lands on library view, not source view; `currentArtistId` syncs so Enhanced toggle works
- [ ] **Smart back button** — open artist-detail from Search / Discover / Watchlist / Stats / Downloads → back button label and route match origin
- [ ] **`/artists` URL** redirects to `/search`
- [ ] **Streaming fallback** — Popular on Last.fm play button streams a track that isn't in the local library
- [ ] **Watchlist config GET** returns enriched artist data without `no such column` errors in logs
- [ ] **Global search widget** still works unchanged (intentionally deferred revamp — see follow-up)
- [ ] Spot-check Sync, Library, Wishlist, Discover, Automations, Stats pages for console errors after navigation changes


core/artist_source_lookup.py
```diff
@@ -0,0 +1,88 @@
...
+    "hydrabase": "soul_id",
+    "musicbrainz": "musicbrainz_id",
+}
+
+
+def find_library_artist_for_source(
+    database,
+    source: str,
+    source_artist_id: str,
+    artist_name: str,
```
kettui: should be marked as Optional to avoid having to pass it in as an empty string if otherwise not available

web_server.py
```diff
@@ -11218,19 +11316,232 @@ def test_artist_endpoint(artist_id):
...
+    try:
+        active_server = config_manager.get_active_media_server()
+    except Exception:
+        active_server = None
+    return _core_find_library_artist_for_source(
+        database, source, source_artist_id, artist_name, active_server=active_server
+    )
+
+
+def _build_source_only_artist_detail(artist_id, artist_name, source):
```
JohnBaumb, line 11337: If we're going to refactor the web_server.py soon, might as well start moving stuff away from web_server.py in our PRs

_build_source_only_artist_detail, make it a module, it's perfect.

JohnBaumb:
The back button stores a single originPage, not a stack. Navigating Search -> Artist A -> Similar Artist B -> Similar Artist C, then pressing back from C jumps straight to Search, skipping A and B.  Probably should fix that in this PR as well.

Nezreka:
> The back button stores a single originPage, not a stack. Navigating Search -> Artist A -> Similar Artist B -> Similar Artist C, then pressing back from C jumps straight to Search, skipping A and B. Probably should fix that in this PR as well.

Good call, that feels much better now. Fix is live in PR

web_server.py
Nezreka: You're right I just pushed a new PR that takes it out of web_server

PR #371, opened 2026-04-24
Feature/search source picker

Nezreka:
Replaces enhanced search's always-fan-out with an explicit source-picker icon row. Each search hits one source; switching sources reads from a per-query cache. Same picker on the Search page and the global Cmd+K widget.

Every keystroke was firing 7 API calls (Spotify, iTunes, Deezer, Discogs, Hydrabase, MusicBrainz, YouTube Videos) regardless of which source the user actually wanted. A single tab could exhaust a Spotify quota in minutes. Now it's 1 call per search.

Also consolidates ~380 lines of near-duplicated source-picker logic from search.js and downloads.js into a single `createSearchController` factory in shared-helpers.js. Net -80 lines across 4 files.

What's here:

- Icon row with real brand logos, glassy chips, brand-color glow on active, pulsing cache dot for already-fetched sources
- Per-query cache — switch sources instantly; new query resets the cache
- Rate-limit fallback banner when backend serves a different source than the one clicked
- Unconfigured sources render dimmed; clicking them opens Settings → Connections
- Global search bar gets an ambient accent aura so it doesn't disappear on dark backgrounds
- Library empty-state offers a "search online for X" handoff to /search
- Mobile responsive
- Soulseek icon → basic-section file results on Search page, /search hand-off on global widget

Drive-by fix: the hourly `clean_search_history` automation was crashing with `'DownloadOrchestrator' object has no attribute 'base_url'`. Guard predates the orchestrator refactor. Fixed to match the pattern at the two other call sites.

Testing: pytest 395/395 green. Manual check across source icons, cache dots, fallback banner, Soulseek hand-off, and library CTA on both surfaces.


webui/static/downloads.js
```diff
@@ -5143,113 +5177,82 @@ function _gsShowResults() {
     if (r && r.innerHTML.trim()) r.classList.add('visible');
 }
 
-async function _gsPerformSearch(query) {
-    if (_gsState.abortCtrl) _gsState.abortCtrl.abort();
-    if (_gsState.altAbortCtrl) _gsState.altAbortCtrl.abort();
-    _gsState.abortCtrl = new AbortController();
-    _gsState.altAbortCtrl = new AbortController();
+function _gsNavigateToSearchPage(query, src) {
```
kettui, line 5180: AI review finding:
bug: Soulseek handoff always writes to `enhanced-search-input`, so the Search page runs the metadata flow instead of raw-file search. Prefill `downloads-search-input` or pass `soulseek` into the Search controller before dispatching input.

webui/static/shared-helpers.js
```diff
@@ -12,6 +12,548 @@
...
+        if (state.sources[src]) {
+            _notify();
+        } else if (state.query) {
+            _fetchSource(src);
+        } else {
+            _notify();
+        }
+    }
+
+    async function _fetchSource(src) {
```
kettui, line 321: Another AI review finding:
`webui/static/shared-helpers.js:L304-L344, L385-L406`: 🟡 risk: stale in-flight requests can still clear `loadingSources` after a fast retype, so the new query can flash empty/error while its fetch is still running. Add a per-query request id/token and ignore completions that do not match before mutating shared state.

webui/static/shared-helpers.js
```diff
@@ -12,6 +12,548 @@
...
+// Canonical display order for the source picker. Standard metadata sources
+// first, then YouTube Music Videos, then Soulseek (basic-file source).
+const SOURCE_ORDER = [
+    'spotify', 'itunes', 'deezer', 'discogs', 'hydrabase', 'musicbrainz',
+    'youtube_videos', 'soulseek',
+];
+
+// Sources the config-status endpoint doesn't cover because they don't need
+// user-supplied credentials — they always render as "configured" in the picker.
+const _ALWAYS_CONFIGURED_SOURCES = new Set(['musicbrainz', 'youtube_videos', 'soulseek']);
```
kettui: Soulseek availability could be checked instead of always having it enabled - I don't have it set up on my dev instance, and it lets me attempt searches even though they are never going to succeed. Ultimately quite a minor thing, but still something that can be prevented

kettui (approved):

PR #374, opened 2026-04-24
Fix Album Completeness job reporting zero findings for every album

Nezreka:
Reported by sassmastawillis: the Album Completeness maintenance job scans 3127 albums in 0.1 seconds and reports 0 findings — for every user, regardless of whether their library is actually complete. Restoring an older DB surfaced 7 correct findings, so the code logic works; the DB state is what's making everything look complete.

Root cause: `albums.track_count` is only ever written by server-sync paths — Plex's `leafCount`/`childCount` and SoulSync standalone's `len(tracks)`. It's the OBSERVED count of tracks SoulSync has indexed, which is always exactly what `COUNT(tracks)` returns for that album. The completeness job treated it as the EXPECTED total and compared it against the observed count. They're equal by construction, so `actual >= expected` is always true: skip, 0.1s scan, 0 findings.

Fix: new `api_track_count INTEGER` column on `albums`, written only by metadata-source code paths. Populated in two places so the scan is fast and the fallback is robust.

1. Enrichment workers — shared helper `set_album_api_track_count` in `core/worker_utils.py`. Called by each worker's existing `_update_album` method alongside its other album-column UPDATEs:

   - spotify_worker: `album_obj.total_tracks` from the Spotify Album dataclass (already in hand, zero new API calls)
   - itunes_worker: same, from the iTunes Album dataclass
   - deezer_worker: `nb_tracks` from full_data, falling back to search_data when the full lookup didn't run
   - discogs_worker: count of tracklist rows where `type_=='track'` (Discogs tracklists interleave heading and index rows that shouldn't count as songs)

   Helper skips the write on zero/None/negative/non-numeric inputs
   so a source lacking track info can't clobber a good value a
   different source already wrote. Caller owns the transaction —
   helper just queues an UPDATE on the caller's cursor without
   committing, so it batches cleanly with each worker's existing
   multi-UPDATE pattern.

   Hydrabase worker deliberately not touched — it's a P2P mirror
   that doesn't write album metadata to the local DB. Hydrabase-
   primary users hit the fallback path below.

2. Album Completeness repair job — new `al.api_track_count` column in the SELECT, read first in the scan loop. On miss (album never enriched, or enrichment workers haven't run yet on a fresh install), falls through to the existing `_get_expected_total()` API lookup and persists the result via the same shared helper (wrapped in connection/commit management since the repair job runs outside a worker's batched transaction).

Also removed `al.track_count` from the scan's SELECT — now unused since the observed count was the whole source of this bug, and leaving a dead SELECT would invite a future engineer to re-introduce the same comparison.

Help text on the job card was reworded so it honestly describes current behavior ("counts cached during normal enrichment are used when available; otherwise the job queries a metadata source directly") rather than the old "active provider first, then others as fallback" phrasing, which doesn't match how the cache actually fills — any enrichment worker that runs can populate it, and the last writer wins. Document-only follow-up if this edge case ever bites in practice: add a `api_track_count_source` column so the scan can prefer the configured primary source's count over others (e.g. deluxe vs. standard edition mismatches). Not worth the complexity today.

For existing users, the first completeness scan after upgrade is fast to the extent their library is already enriched: the workers already ran and populated `api_track_count` on their normal schedule. For brand-new installs, the scan's fallback path handles the cold start — slower, but correct, and subsequent scans are fast.

Does NOT affect:
- Download / post-processing / wishlist / sync code paths — none of them read `track_count` for completeness semantics.
- Plex / Jellyfin / Navidrome / standalone sync — still write `track_count` exactly as before; `api_track_count` is a separate column they never touch.
- Other repair jobs.
- Any UI path — same finding schema, just correct counts now.

Files:
- database/music_database.py — idempotent migration adding `api_track_count INTEGER DEFAULT NULL` to the existing album-column check block.
- core/worker_utils.py — new `set_album_api_track_count` helper with the documented skip-on-bad-input contract.
- core/spotify_worker.py, itunes_worker.py, deezer_worker.py, discogs_worker.py — one-liner call from each `_update_album`.
- core/repair_jobs/album_completeness.py — scan uses the cache; fallback path persists API-lookup results via the shared helper; help text updated to match actual behavior.
- tests/test_worker_utils_album_track_count.py — 9 tests covering the helper's write/skip contract + no-commit invariant.
- tests/test_album_completeness_job.py — 2 tests for the repair job's fallback-path wrapper.
- webui/static/helper.js — WHATS_NEW entry.

Credit: sassmastawillis spotted the bug; the "restored older DB finds 7 albums" signal pinpointed DB state over code logic and made the diagnosis tractable.


database/music_database.py
```diff
@@ -3122,6 +3122,19 @@ def _add_soul_id_columns(self, cursor):
...
+            # Albums: api_track_count — cached expected track count from the
+            # metadata provider, separate from track_count which is the
+            # OBSERVED count written by server syncs (Plex leafCount,
+            # SoulSync standalone len(tracks)). Without a separate column,
+            # the Album Completeness job can't tell apart "you have all the
+            # tracks" from "Plex says this album has N tracks and you have
+            # N tracks" — the latter looks complete but might be missing
+            # material the metadata source knows about. NULL = not yet
+            # looked up; the repair job fills it as it runs.
+            if 'api_track_count' not in album_cols:
```
kettui, line 3134: > database/music_database.py:L4724-L4731: 🟡 risk: api_track_count is not copied during the ratingKey migration, so the cache disappears when an album row is rekeyed. Add it to enrichment_cols or the next completeness scan will fall back to live API lookups again.

core/discogs_worker.py
```diff
@@ -454,6 +454,18 @@ def _update_album(self, album_id, data: Dict[str, Any]):
...
+            # Cache the authoritative expected track count for the Album
+            # Completeness repair job. Discogs tracklists mix actual tracks
+            # with section headings and index entries — only type_=='track'
+            # rows are real songs, so filter before counting. (Helper can't
+            # do this filter for us; it's Discogs-specific shape.)
+            tracklist = data.get('tracklist') or []
+            set_album_api_track_count(
+                cursor,
+                album_id,
+                sum(1 for t in tracklist if t.get('type_') == 'track'),
```
kettui: > core/discogs_worker.py:L462-L466: 🟡 risk: Discogs counting is too strict. `get_album_tracks()` keeps `type_ == ''` as real tracks too, so this undercounts releases with blank `type_` entries. Count `type_ in ('track', '')` or use the normalized track list instead.

Nezreka: Okay I just pushed the fix for both of those issues as well as the merge conflicts.

PR #377, opened 2026-04-25
Reorganize: route library files through the post-processing pipeline

Nezreka:
Reported on Discord by winecountrygames. The library "Reorganize" tool had several layered bugs that all traced to the same root cause: the endpoint reinvented every wheel post-processing already turns — its own template engine, its own disc-number resolution from file tags, its own sidecar sweep, its own collision detection — and each had drifted from the canonical path used by fresh downloads. Reported symptoms:

  - 3-disc Aerosmith deluxe collapsed to a flat single-disc layout
  - Half the tracks on other albums silently skipped, no error / no count
  - Re-runs left empty leftover album folders cluttering the artist dir

Architecture: stop reinventing wheels. Route reorganize through exactly the same pipeline downloads use. Per-album:

  1. Fetch the canonical tracklist from a metadata source (Spotify / iTunes / Deezer / Discogs / Hydrabase) using the album's stored source IDs. New `core/library_reorganize.py::plan_album_reorganize` does this — primary-source-first, fall through priority chain unless the user picked a specific source in the modal (strict mode).
  2. For each local track, find the matching API entry via a scored candidate matcher. Score components: exact-title (100), substring-with-length-ratio (40-90), track-number agreement (20). Hard reject when the two titles have different version differentiators (Remix vs no-remix means different recordings, not annotation drift). Below threshold = unmatched, surfaced as "not in source's tracklist, left in place" rather than silently mis-routing.
  3. Copy the file to a per-album staging directory, build the same context dict the import flow builds (`spotify_album` / `track_info` / etc. with `is_album_download=True` so the path builder enters ALBUM mode, not SINGLE mode), call `_post_process_matched_download(...)` — same function fresh downloads use. Post-process handles tagging, multi-disc subfolder decisions, sidecar regeneration, AcoustID verification.
  4. Read `context['_final_processed_path']` to learn where it landed. Update `tracks.file_path` in the DB BEFORE removing the original (DB-update failure leaves the file at both locations, recoverable via library scan; the reverse would orphan the row). Delete per-track sidecars (post-process recreates them at the new destination).

3 concurrent workers per album via ThreadPoolExecutor, matching the download path's per-batch worker count. State mutations all guarded by a single lock; staging filenames carry a UUID prefix so concurrent copies of identically-named source files don't overwrite each other.

Source picker in the modal lets the user choose which source to read the tracklist from. Two endpoints feed it:
  - `/api/library/album/<id>/reorganize/sources` — sources for THIS album that are both authed AND have a stored ID. For the per- album modal.
  - `/api/library/reorganize/sources` — all authed sources globally. For the bulk "Reorganize All" modal where per-album ID coverage varies. When the user picks a specific source, the orchestrator runs in `strict_source=True` mode (no fallback chain) — picking Spotify means "use Spotify or fail", not "use Spotify and silently fall back."

Preview endpoint shares the same planning logic as apply via `preview_album_reorganize` — the destination path comes from the same `_build_final_path_for_track` post-process uses, so what you see in the preview is exactly what you get on apply.

Empty destination folders (from earlier failed runs OR from the current run when post-process creates a dir then fails AcoustID) get cleaned up after each successful run: walk up to the artist folder from any successful destination, prune empty album-sibling folders one level deep. Bounded scope = won't touch unrelated user dirs.

Web_server.py shrinks by ~450 net lines. The endpoint handler is now a thin wrapper that builds injected callables (path resolver, post- process function, DB updater, empty-dir cleaner), spawns a thread that calls `reorganize_album()`, and returns. All actual logic lives in `core/library_reorganize.py` where it's unit-testable without spinning up Flask.

Frontend cleanup: the per-call template input in both reorganize modals (per-album and bulk) was redundant — the backend always uses the configured global download template. Removed the input and the variables-grid reference UI it was for.

39 new unit tests pin every contract:
  - source resolution (no_source_id when album has none, fallthrough chain when primary returns nothing, strict mode bypasses fallback)
  - matcher scoring (exact / substring / multi-disc disambiguation / smart-quote tolerance / dash-vs-parens / bonus-track substring / Remix-vs-original differentiator rejection / "Real" doesn't false- match "Real Real Real" / track-number-only no longer fires)
  - file safety (DB-update failure leaves original in place, post- process failure leaves original in place, post-process exception caught and original preserved, success removes original AND updates DB in the right order)
  - sidecar handling (per-track .lrc/.nfo deleted on success, kept on failure; album-level cover.jpg/folder.jpg cleaned only when directory has no remaining audio)
  - staging cleanup (recreated between tracks because post-process nukes it, dir cleaned up on success AND on failure)
  - destination-dir prune (empty siblings removed, real album with files preserved, no recursive sweep)
  - source picker (only authed-with-stored-ID sources for per-album, all authed sources for bulk; strict mode doesn't fall back)
  - concurrency (3 workers in flight, state stays consistent under races, stop_check cuts off pending tasks)
  - preview parity (preview produces same destination as apply for multi-disc; ALBUM mode not SINGLE mode; unmatched/no-path tracks surfaced with reasons)

Limitations (deliberate punts, NOT in this PR):
  - Renamed local titles on multi-disc albums where track_number also disagrees: matcher returns nothing (track is "not in source"). Fixable by using duration_ms as a tertiary signal.
  - Per-track in-modal source switching with per-album track-count hints (would need a second API call before opening the modal).
  - UI status panel on the artist page during a run — currently just toasts. Documented as a follow-up PR.

Files:
  - core/library_reorganize.py — new module: plan_album_reorganize, preview_album_reorganize, reorganize_album, available_sources_for_album, authed_sources, _score_candidate, helpers for staging/post- processing/finalizing, sidecar + dest-dir cleanup
  - core/metadata_service.py — no changes; reused get_album_for_source, get_album_tracks_for_source, get_source_priority, get_client_for_source
  - web_server.py — three endpoints (preview / apply / sources GETs) are thin wrappers; -450 net lines
  - tests/test_library_reorganize_orchestrator.py — 39 tests covering every contract above
  - webui/static/library.js — source picker UI in both modals; dead template input + variables-grid removed
  - webui/static/style.css — dropdown option styling fix (white-on- white was unreadable)

This touches web_server.py but most removing things so i'm happy to wait on this PR until you finish breaking up the import function from web_server.py


core/library_reorganize.py
```diff
@@ -0,0 +1,1477 @@
...
+
+    staging_file = _stage_track(ctx, track_id, title, resolved_src)
+    if staging_file is None:
+        return
+
+    new_path = _run_post_process_for_track(ctx, track_id, title, plan_item['api_track'], staging_file)
+    if new_path is None:
+        return
+
+    _finalize_track(ctx, track_id, resolved_src, new_path)
```
kettui: core/library_reorganize.py:L1088-L1094: 🔴 bug: `_finalize_track()` can bail on DB update failure, but `moved` is still incremented unconditionally. Return a success flag and only count the track as moved after the row update and source removal actually happened.

webui/static/library.js
kettui: webui/static/library.js:L6312-L6314: 🔴 bug: `result.total` is undefined now that `/reorganize` no longer returns it, so the toast says "Reorganizing undefined tracks...". Return `total` again or switch this toast to `result.message`.

kettui, line 6367: webui/static/library.js:L6337-L6344: 🟡 risk: every `done` run is toasted as success unless `failed > 0`, so `no_source_id` and other zero-failure skips look green. Carry `state.status` through and warn on non-completed runs or any `state.errors`.

webui/static/library.js
```diff
@@ -6476,7 +6476,7 @@ async function _executeReorganizeAll() {
             const resp = await fetch(`/api/library/album/${album.id}/reorganize`, {
```
kettui, line 6509: webui/static/library.js:L6475-L6490: 🟡 risk: bulk mode treats any completed request as succeeded without checking the final `/status` payload, so albums skipped for missing source IDs are counted as successes. Gate `succeeded` on the finished state instead.

Nezreka: Okay fix is live for all four issues

kettui (approved):

Issue #369, opened 2026-04-23
Silent exception swallowing hides bugs and failures


JohnBaumb:
## Problem

There are 15+ instances of `except Exception: pass` in `web_server.py` alone, with more across the codebase in `api/system.py`, `api/search.py`, `services/sync_service.py`, and `beatport_unified_scraper.py`.

Examples from `web_server.py`:

```python
except Exception:
    pass  # DB error -- don't block requests, use the session value

except Exception:
    pass  # Non-critical, never block download flow

except Exception:
    pass
```

Silent exception swallowing means:

- Bugs are invisible -- errors happen but nobody knows
- Debugging production issues becomes guesswork since the error context is lost
- "Non-critical" failures can cascade into critical ones (e.g., a silently failed DB write means stale data is served forever)
- The application can enter inconsistent states without any log trace

### Replace pass with logging

Every `except Exception: pass` should at minimum log at WARNING or DEBUG level

Nezreka:
Cleaned up!

PR #378, opened 2026-04-25
Refactor import and related post-processing pipelines

kettui:
## Summary
- Break the monolithic import flow out of `web_server.py` into dedicated modules under `core/imports`.
- Move shared metadata responsibilities into `core/metadata` and tighten `core/metadata_service.py` into a thinner orchestration layer.
- Make single-track import resolution source-aware so imported tracks keep the correct artist and album metadata.
- Update the UI import flow to pass manual single-match selections through correctly.
- Clean up source-aware logging and naming in the import/wishlist path.

## What Changed
- Split import context, path building, staging, resolution, side effects, guards, and post-processing into separate modules.
- Extracted metadata enrichment, artwork, lyrics, and source helpers into `core/metadata`.
- Added source-detail and source-artist lookup helpers to support source-only artists cleanly.
- Updated the single-import flow to use the selected manual match and preserve inferred metadata.
- Refreshed import and metadata tests to cover the new module boundaries and behaviors.
- Updated the frontend search/import UI and shared helpers to align with the new source-aware structure.

## Validation
- Added and updated focused tests across `tests/imports` and `tests/metadata`.
- Verified the split modules and key import paths with targeted sanity checks.
- Manually imported a single and an album

## Notes
- This is a broad refactor with some compatibility-focused cleanup still left in place.
- A few source-specific field names remain in storage and legacy paths for now to avoid a larger schema migration in this PR.
- Existing metadata files such as metadata_service and metadata_cache were not moved to the new metadata package with the goal of not widening the scope of the PR even further. Those can be done in future PR's
- Some import / post-processing related logic is still Spotify-driven (eg. wishlist). Those need to be refactored in a separate PR as well.
- Manual imports that call the `/api/import/*/process` endpoints are still driven by the UI, so those require the user to stay on the site when importing things. This processing should ideally be moved to a background worker in a future PR.


kettui:
Can be tested already, but still ironing things out.

PS. the web_server.py reductions turned out to only be 5k lines, so luckily not the 20k that it initially seemed out to be

web_server.py
```diff
@@ -471,6 +527,172 @@ def _make_context_key(username, filename):
 
 logger.info("Core service initialization complete.")
 
+# --- Shared Runtime State ---
```
kettui, line 655: These state declarations were a bit further down originally (so these are not new additions), but they moved up as a result of some technical mishaps. Everything should be intact, though

Nezreka:
A quick claude review on the diff:


## 🔴 Bugs

### `tests/imports/test_import_paths.py:72` — fails on Windows

```
AssertionError: assert 'Artist One\\Album One' == 'Artist One/Album One'
```

The test expects `'Artist One/Album One'` but `get_file_path_from_template_raw` returns `'Artist One\\Album One'` on Windows. Implementation uses `os.path.join` (or `Path` ops) which produces backslashes on Windows, forward slashes on Linux. CI likely runs on Linux so this is masked there — Windows users hit it on every test run.

**Fix:** either normalize the implementation's output to forward slashes (web-safe / cross-platform — file paths on disk and template-rendered paths can sensibly be different), or compare against `os.sep` in the test.

### `core/metadata/source.py:228-235` — mb_release_cache poisons on exception

```python
try:
    rc_result = mb_service.match_release(album_name_for_mb, artist_name)
    pp["release_mbid"] = rc_result.get("mbid", "") if rc_result else ""
except Exception:
    pp["release_mbid"] = ""
mb_release_cache[rc_key_norm] = pp["release_mbid"]
mb_release_cache[rc_key_exact] = pp["release_mbid"]
```

When `match_release` raises (transient MusicBrainz outage, rate limit, network blip), the exception is caught and `""` is cached for both keys. Every subsequent track in the same artist/album lookup hits the cache, sees the empty string, and skips the MusicBrainz call forever — even after MB recovers. Same anti-pattern that came up on the reorganize DB helpers earlier.

**Fix:** only cache non-empty results. A failed lookup should not populate the cache at all (or use a sentinel that's distinguishable from "successfully looked up and got nothing").

## 🟡 Risks

### Inline `from core.metadata_service import ...` inside function bodies

`core/imports/staging.py:51,57,63,134`, `album.py:15,26`, `resolution.py:14`, several others. Smells like a circular-import workaround — `core/metadata_service.py` likely imports from `core/imports/` too. The lazy-import dodges the cycle but the underlying coupling is still there.

If this is genuinely a circular dependency, the design fix is to split `metadata_service.py` into a smaller core that both `core/metadata_service.py` and `core/imports/*` can depend on without back-references. Or accept the lazy imports but document them with a single-line comment explaining the cycle being broken.

### Inline `from config.settings import config_manager` inside fn bodies

`core/imports/file_ops.py:16`, `guards.py:28`, `side_effects.py:34`. `config_manager` has no circular dep relationship with these modules — there's no reason for the lazy import. Move to module top.

### `core/runtime_state.py:54-67` — `mark_task_completed` lock contract

Was changed in `1820e33` from internal-acquire to caller-must-hold (to avoid double-acquire — defensible). But the discipline-only contract is fragile now that the function is reachable from many call sites in the new package layout. A future caller forgets to acquire `tasks_lock`, mutates `download_tasks` concurrently with another thread, and silently corrupts the dict.

**Options:** assert `tasks_lock.locked()` in dev/debug builds, switch to `RLock` and re-acquire internally (slight overhead but safe), or add a `# CALLER MUST HOLD` decorator that documents and enforces.

### `core/metadata/source.py:41-44` — `mb_release_cache` + `mb_release_detail_cache` unbounded

Two module-level dicts, RLock-guarded, no upper bound. Long-running library scans grow these into hundreds of thousands of entries. Migration-faithful (probably matched the inline shape that was in `web_server.py`) but the refactor was the moment to fix it.

**Fix:** wrap with `cachetools.LRUCache` or similar with a sane cap (~10k). Existing RLock pairs migrate cleanly.

### `core/metadata/common.py` — `_FILE_LOCKS` cache unbounded

Same long-lived-server problem. Every distinct file path inserts a `threading.Lock` and never removes it.

**Fix:** `WeakValueDictionary`, or LRU-evict when the dict exceeds N entries.

### `core/metadata_service.py:139-145, 1985-2010` — `importlib.import_module('web_server')` at use-site

The Spotify and Hydrabase branches in `get_client_for_source` reach back into `web_server.py` via runtime `importlib` calls to fetch the live client. This recreates the very `web_server`↔`core` coupling the refactor exists to break, makes the function untestable without a fully-loaded `web_server`, and means `clear_cached_metadata_clients()` doesn't actually clear Spotify/Hydrabase.

**Fix:** add `register_spotify_client(client)` / `register_hydrabase_client(client)` setters in `metadata_service`. Have `web_server` call them once at startup. Pure DI.

## ⚪ Nit

### `core/metadata/source.py` per-source error swallow

Every `_process_*_source` block wraps everything in `except Exception as exc: logger.error("X failed (non-fatal): %s", exc)`. This catches `TypeError`, `AttributeError`, `KeyError` — programmer errors that shouldn't be silenced — and logs them as "non-fatal", which means they never get noticed and never get fixed. Same anti-pattern as #369.

**Fix:** catch a narrow tuple (`requests.RequestException`, `socket.timeout`, etc.) for network calls. Let programmer errors raise.

## Summary

The two reds (Windows test failure + mb_release_cache poisoning) are the merge-blockers from my read. The `importlib` web_server lookup is the single biggest architectural smell — it half-undoes the refactor's testability win — and worth fixing in this PR rather than a follow-up. The unbounded caches and inline imports can defer to later cleanups.

Otherwise: solid extraction, clean separation, good test coverage. Architecture is the right shape.

kettui:
I'll get around to checking these out later today (EU time) or tomorrow morning.
I'll quickly comment on this though

> Windows users hit it on every test run

I personally hate the idea of explicitly supporting Windows as a development platform (I don't see a reason not to use WSL if you're developing on Windows), but I understand you use Windows yourself @Nezreka (?)

I'll make sure to handle it here, but in any case this can (and will) cause problems in the future if Windows support is a priority, since most devs don't have a Windows setup at their disposal for ensuring that things work there also. Linux, macOS and WSL are easy to support without much of an issue, but keeping "raw" Windows in the loop makes things difficult, imo.

Nezreka:
Okay we can move to wsl development. I didn't know this was a pain point. ill look into what it takes now

kettui:
> the design fix is to split metadata_service.py into a smaller core that both core/metadata_service.py and core/imports/* can depend on without back-references.

It's definitely a goal to do further splitting for the metadata_service in the future, as there currently is a natural place to do such splitting through the added `core/metadata` package. It was left somewhat untouched in this PR due to the fact that a lot of things import it currently, so it would've blown the scope out of proportions if refactored in the same go

Nezreka:
378 is good. all issues that are there are now were already there before your changes. so much easier to isolate now though.

PR #379, opened 2026-04-26
Library reorganize: FIFO queue with live status panel

Nezreka:
Replaces the single-slot "one reorganize at a time, return 409 on collision" model with a per-user FIFO queue. Buttons stay clickable, "Reorganize All" is one backend call instead of an N-call JS loop, and a status panel mounted at the top of the artist actions bar shows live progress (active item, queued count, recent completions) with per-item cancel buttons.

Backend
- core/reorganize_queue.py: singleton queue + worker thread, dedupe-on- enqueue, cancel rules (queued cancellable, running not), enqueue_many for bulk operations, progress fan-out via update_active_progress
- core/reorganize_runner.py: factory builds the worker's runner closure with injected dependencies. Reads config per-call so changing the download path in Settings takes effect on the next reorganize without a server restart
- database/music_database.py: get_album_display_meta and get_artist_albums_for_reorganize — moves the SQL out of route handlers
- web_server.py: thin enqueue/snapshot/cancel/clear endpoints, runner registration at module load. Old _reorganize_state globals + status endpoint deleted. Static-asset cache buster (?v=<server-start>) added so JS/CSS updates ship live without users clearing cache

Frontend
- webui/static/library.js: status panel mount, polling (1.5s when active, 8s when idle), expand/collapse, per-item cancel, debounced enhanced-view reload (one reload per artist batch instead of N). Per-album reorganize button paints with queued/running indicator and short-circuits to a toast when the album is already in queue
- webui/static/style.css: panel + button styling matching the existing glass-UI accents
- webui/static/helper.js + version modal: WHATS_NEW entry

Tests (22 new)
- tests/test_reorganize_queue.py (19 tests): FIFO order, dedupe, per-item source, cancel rules, continue-on-failure, snapshot shape, progress propagation, bulk enqueue
- tests/test_reorganize_runner.py (4 tests): per-call config reads, setup-failure summary, dependency injection, progress fan-out
- tests/test_reorganize_db_methods.py (7 tests): SQL JOIN behavior, ordering, fallback for blank strings, artist isolation

Full suite 549 passed in 27s.


webui/static/library.js
kettui, line 6222: webui/static/library.js:L6224: 🟡 risk: preview errors leave #reorganize-apply-btn disabled because the early return skips the re-enable path. Move the button reset into finally or re-enable before every return.

core/reorganize_queue.py
```diff
@@ -0,0 +1,413 @@
...
+                if item.status == 'queued':
+                    return item
+            return None
+
+    def _run(self) -> None:
+        """Worker loop: pull next queued, run it, mark done, repeat.
+        Idles on `_wakeup` event when queue is empty."""
+        logger.info("[Queue] Worker thread started")
+        while not self._stopped:
+            item = self._next_queued()
```
kettui: core/reorganize_queue.py:L316: 🔴 bug: _wakeup.clear() can race with enqueue()'s _wakeup.set(), so a newly queued album can sleep until the 60s timeout. Use a Condition or re-check the queue under the same lock before clearing.

core/reorganize_queue.py
```diff
@@ -0,0 +1,413 @@
...
+
+        Returns a tally dict ``{'enqueued': N, 'already_queued': M,
+        'total': len(items)}`` so the caller can report bulk results
+        without doing the counting themselves. Used by the bulk
+        Reorganize-All endpoint and any future maintenance jobs that
+        enqueue at scale.
+        """
+        enqueued = 0
+        already = 0
+        for item in items:
```
kettui: core/reorganize_queue.py:L210: 🟡 risk: enqueue_many() dedupes item-by-item, so a duplicate later in the same bulk request can slip through if the first copy finishes before the loop reaches it. Deduplicate the batch up front or append the whole batch before starting the worker.

core/reorganize_queue.py
kettui: core/reorganize_queue.py:L316-L326: 🔴 bug: a queued item can be canceled after _next_queued() picks it but before status flips to running, then the worker still runs it. Recheck item.status under the lock before starting, or mark it running atomically when selecting it.

database/music_database.py
```diff
@@ -4795,6 +4795,69 @@ def insert_or_update_media_album(self, album_obj, artist_id: str, server_source:
...
+                )
+                row = cursor.fetchone()
+                if not row:
+                    return None
+                return {
+                    'album_title': row['album_title'] or 'Unknown Album',
+                    'artist_id': str(row['artist_id']) if row['artist_id'] is not None else None,
+                    'artist_name': row['artist_name'] or 'Unknown Artist',
+                }
+        except Exception as e:
```
kettui: > database/music_database.py:L4828-L4859: 🟡 risk: both new DB helpers swallow query failures and return None/[], so a real DB outage shows up as “Album not found” / “No albums found” instead of a 500. Let the exception bubble or return an explicit error sentinel.

Most db query methods in the codebase currently return empty responses like this in the case of a failure, so probably wouldn't worry too much about this, unless we want the failure to affect the outcome of the feature that uses it. In this case db errors appear to the user as "album not found", which probably isn't ideal, but isn't the absolute worst thing either, as someone would likely create an issue / ask about it in Discord if a seemingly false positive like that were to happen

kettui: In general though I don't see a reason not to do things "properly" with any new / refactored feature that gets written. Proper error-handling can be a struggle, but it's ultimately very important as the features keep evolving as fast as they have been in this codebase

Nezreka: excellent review. I'll remember these tips. fixing now

Nezreka: good catch — both helpers now let exceptions bubble. callers in web_server.py already 500 on raise, so a real DB outage shows up properly instead of "album not found". added two regression tests.

agreed — applying that to anything new from here on. thanks for the nudge.

core/reorganize_queue.py
Nezreka: real bug — fixed. pick + flip-to-running is now atomic under one lock acquisition. added a regression test that hammers enqueue+cancel pairs and asserts cancelled items never run.

core/reorganize_queue.py
Nezreka: ooof yeah. enqueue_many now holds the lock for the whole batch and tracks a per-batch seen set, so duplicates inside one bulk call dedupe against each other too. test added.

core/reorganize_queue.py
Nezreka: yep — swapped the lock + event for a single threading.Condition. cond.wait() releases and re-acquires atomically on notify, so no missed wakeups.

webui/static/library.js
Nezreka: fixed — moved the re-enable into a finally with a canApply flag. every exit path now lands the button right.

Nezreka:
thanks @kettui — fixed all five. pushed with 3 regression tests.

PR #384, opened 2026-04-27
Pin yt-dlp in requirements.txt, drop pip install from entrypoint

### LINKED ISSUE #367: Entrypoint runs pip install on every container start
## Problem

`entrypoint.sh` runs `pip install -U yt-dlp` unconditionally on every container start:

```bash
pip install -U yt-dlp --quiet --no-cache-dir 2>/dev/null && echo "   yt-dlp updated" || echo "   yt-dlp update failed (will use existing version)"
```

### Option A: Pin in requirements.txt with periodic bumps

Keep yt-dlp pinned in `requirements.txt` (it already is with `>=2024.12.13`). Change the pin to an exact version and update it via PR/CI when new releases come out. Remove the entrypoint pip install entirely.

### Option B: Scheduled update instead of startup update

If auto-updating yt-dlp is a product requirement, move it to a background task that runs once per day (via the existing automation engine) instead of blocking container startup. This keeps startup fast and predictable.

### Option C: Keep it but add hash verification

If the entrypoint install must stay, pin to a specific version with `--hash`:

```bash
pip install 'yt-dlp==2025.01.15' --require-hashes --no-cache-dir 2>/dev/null
```

This is the least disruptive change but requires updating the hash on each version bump.

Nezreka:
Closes #367 (reported by JohnBaumb).

The Docker entrypoint ran `pip install -U yt-dlp --quiet --no-cache-dir` on every container start. Three problems with that:

- Non-deterministic startup: each restart could pick up a different yt-dlp version, making "works on my machine" debugging harder.
- Network dependency at boot: PyPI being slow/unreachable gated the app coming up.
- In-place upgrades inside running containers can race with active yt-dlp invocations and aren't a great pattern.

Picked Option A from the issue: pin to an exact version in requirements.txt (`yt-dlp==2026.3.17`) and remove the entrypoint install entirely. yt-dlp comes baked into the image now via the existing `pip install -r requirements.txt` in the Dockerfile.

Tradeoff: YouTube fixes ship via SoulSync releases now instead of "next container restart". The pin is documented inline with how to bump it.

Net change: -3 entrypoint lines, requirements.txt pin tightened, WHATS_NEW '2.4.1' block opened (entries hidden until version bumps).

553 tests pass.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,7 +28,12 @@ beautifulsoup4>=4.12.0
 psutil>=6.0.0
 
 # YouTube support
-yt-dlp>=2024.12.13
+# Pinned to an exact version for reproducible builds. The Docker
+# entrypoint used to `pip install -U yt-dlp` on every container start,
+# which made startup non-deterministic and depended on PyPI being
+# reachable. Bump this pin in a SoulSync release when yt-dlp ships a
+# YouTube-fix update — that's the supported upgrade path now.
+yt-dlp==2026.3.17
 
 # Lyrics support
 lrclibapi>=0.3.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -27,12 +27,7 @@ beautifulsoup4>=4.12.0
 # System monitoring
 psutil>=6.0.0
 
-# YouTube support
-# Pinned to an exact version for reproducible builds. The Docker
-# entrypoint used to `pip install -U yt-dlp` on every container start,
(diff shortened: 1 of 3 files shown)


requirements.txt
Nezreka, line 1: risk: pin `2026.3.17` is March release, today is April 26. Likely newer yt-dlp out. Bump to current latest before merge or doc pin-pick policy.

PR #388, opened 2026-04-27
Lay the groundwork for webui React transition

kettui:
## Summary
- Lay the groundwork for the React webui migration while keeping the legacy shell in place.
- Move `/issues` into React as the first migrated route and use it as the reference implementation for future page slices.
- Update the dev/build/test pipeline so the new frontend can coexist cleanly with the old shell during the transition.

## What Changed
- Added the Vite/React app scaffold under `webui/src/app/`, including TanStack Router, React Query, the shared query client, and the browser shell bridge.
- Migrated the `issues` route into React, including the filtered issue list, counts, detail modal, admin mutations, and the legacy report-issue handoff.
- Split the route into clear modules:
  - `route.tsx`
  - `-issues.types.ts`
  - `-issues.api.ts`
  - `-issues.helpers.ts`
  - `-ui/`
- Added shared UI primitives and route-local form handling so the new page does not depend on the legacy shell widgets.
- Kept the legacy shell aligned with React navigation, profile selection, browser history, and route gating while the rest of the app stays on the old runtime.
- Centralized shell bridge glue and updated the shell load order so the legacy JS shell and React app can boot together safely.
- Updated the Docker build, and Flask asset pipeline to support the new frontend workflow.
- Added a convenience `dev.py` script that starts the python server as well as vite server in dev mode.
- Added documentation for the migration analysis and the new Issues route architecture so future pages can follow the same pattern.
- Added tests for router behavior, issue helpers, request-layer behavior, and a Playwright smoke path for the migrated route.

## Notes
- This is an incremental migration, not a full webui rewrite.
- Most pages still render through the legacy shell.
- `/issues` is the first React-owned route and the pattern to copy for future migrations.
- Request-level tests now use MSW so API success and error paths stay realistic and easy to extend.


webui/static/library.js
```diff
@@ -700,6 +700,27 @@ let artistDetailPageState = {
...
+    }
+
+    artistDetailPageState.currentArtistId = null;
+    artistDetailPageState.currentArtistName = null;
+    artistDetailPageState.currentArtistSource = null;
+    artistDetailPageState.originStack = [];
+}
+
+if (typeof window !== 'undefined') {
+    window.addEventListener(PAGE_WILL_CHANGE_EVENT, (event) => {
```
kettui, line 716: custom event to clear the current artist details when navigating back/forth from `/artist-detail`. Without this, navigating backwards from artist-detail won't actually change the page

webui/static/shell-bridge.js
```diff
@@ -0,0 +1,174 @@
...
+
+function openAddToWishlistAlbumWorkflow(input) {
+    if (typeof openAddToWishlistModal !== 'function') {
+        throw new Error('Wishlist workflow host is not ready yet');
+    }
+
+    return openAddToWishlistModal(input.album, input.artist, input.tracks, input.albumType);
+}
+
+window.SoulSyncWorkflowActions = {
```
kettui, line 135: this file is where we explicitly expose functionality from the legacy js land to React (even though the React code likely could refer to any global js regardless)

webui/static/stats-automations.js
```diff
@@ -6273,1034 +6273,24 @@ function _autoInsertVar(textareaId, variable) {
...
-    `;
-    return card;
-}
-
-// --- Report Issue Modal ---
-
-let _reportIssueState = {};
+// --- Report Issue Helper ---
 
 function showReportIssueModal(entityType, entityId, entityName, artistName, albumTitle) {
```
kettui, line 6393: `showReportIssueModal` now simply delegates the operation to React via the bridge method `bridge.openReportIssue(...)`

dev.py
```diff
@@ -0,0 +1,348 @@
+#!/usr/bin/env python3
+"""SoulSync development launcher.
```
kettui, line 2: convenience script for running the python backend and vite dev server at the same time

webui/src/routes/issues/-ui/issues-page.tsx
```diff
@@ -0,0 +1,422 @@
...
+
+  const openIssue = (issueId: number) => {
+    void navigate({
+      to: Route.fullPath,
+      search: (prev) => ({ ...prev, issueId }),
+    });
+  };
+
+  const onCategoryChange = (category: IssuesSearch['category']) => {
+    void navigate({
```
kettui, line 84: with React we can easily persist state to (and derive it from) url search params, so that refreshing the page / navigating back and forth between pages can automatically bring us back to the same page state, rather than relying on in-memory state that is managed explicitly via code (like the old vanilla js does)

kettui:
This is quite a beefy PR, but I wanted to make sure a lot of the basic needs are covered from the get-go. The description summarizes it already quite thoroughly so I won't reiterate everything here, but I'll just say that I spent quite a while iterating over this (as the commit log suggests), so I'm quite happy with the end result. This is of course just a start (quite a comprehensive one at that, though), and while there are quite a few conventions and decisions made already in this initial "seed", there will naturally be a lot of missing pieces that get filled in later on as the migration work continues.

core/webui/assets.py
```diff
@@ -0,0 +1,126 @@
...
+) -> str:
+    """Return HTML tags for the WebUI bundle or dev client."""
+    if placement not in ("head", "body"):
+        return ""
+
+    dev_mode = _resolve_dev_mode(dev)
+    vite_url = _resolve_dev_url(dev_url)
+    static_url = static_url_builder or default_static_url_builder
+
+    if dev_mode:
```
kettui, line 101: in dev mode the React assets are served directly from the vite dev server - otherwise the assets are loaded directly similar to the legacy .js files

PR #393, opened 2026-04-28
fix HIFI downloads to get LOSSLESS and HI_RES again

elmerohueso:
Tidal appears to have made a recent change that resulted in the /track endpoint manifests returning HIGH quality when LOSSLESS and HI_RES were requested. This changes to a new workflow using the /trackManifests endpoint, allowing all qualities to work again.


Nezreka:
Hey this looks great, thanks for the PR. All our work flows through dev -> main only at release time. Can you change the base branch on this pr to dev?

elmerohueso:
> Hey this looks great, thanks for the PR. All our work flows through dev -> main only at release time. Can you change the base branch on this pr to dev?

Done

Nezreka:
hey thanks for digging into this. the /trackManifests rewrite + hls + ffmpeg demux looks solid, and you're right that tidal changed the /track endpoint behavior — appreciate the catch.

couple things before i can merge though:

1. the new `/api/hifi/instances` POST/DELETE/reorder/list endpoints need to be admin-only. settings-class routes got locked down in PR #385 because anyone on the local network can hit them otherwise — and these can add malicious instances that route downloads through them. just slap `@admin_only` on the four new routes and we're good there.

2. would you mind adding some tests? at minimum the 6 new db methods (get, get_all, add, remove, reorder, seed) and maybe a few unit tests around `_parse_hls_playlist` for the edge cases (master vs variant playlist, missing init segment, no segments). this is a pretty big rewrite of the download path with zero coverage so it's hard to verify nothing drifts later.

smaller stuff that's not blocking but worth thinking about:
- per-segment retry before bailing the whole download to the next quality tier — one bad segment kills the whole flac
- the DELETE route uses the URL as a path param (`<path:url>`) which can double-encode weird. body or query string is cleaner
- the reorder endpoint doesn't validate the URLs actually exist (silent 0-row update)
- `escapeHtml` is called on the URL in settings.js — just confirm that helper exists in the codebase, otherwise xss vector through a malicious instance URL

ping me once the auth + tests are in and i'll merge it. really good work on the actual fix

elmerohueso:
> hey thanks for digging into this. the /trackManifests rewrite + hls + ffmpeg demux looks solid, and you're right that tidal changed the /track endpoint behavior — appreciate the catch.
> 
> couple things before i can merge though:
> 
> 1. the new `/api/hifi/instances` POST/DELETE/reorder/list endpoints need to be admin-only. settings-class routes got locked down in PR [Gate /api/settings endpoints behind admin profile #385](https://github.com/Nezreka/SoulSync/pull/385) because anyone on the local network can hit them otherwise — and these can add malicious instances that route downloads through them. just slap `@admin_only` on the four new routes and we're good there.
> 2. would you mind adding some tests? at minimum the 6 new db methods (get, get_all, add, remove, reorder, seed) and maybe a few unit tests around `_parse_hls_playlist` for the edge cases (master vs variant playlist, missing init segment, no segments). this is a pretty big rewrite of the download path with zero coverage so it's hard to verify nothing drifts later.
> 
> smaller stuff that's not blocking but worth thinking about:
> 
> * per-segment retry before bailing the whole download to the next quality tier — one bad segment kills the whole flac
> * the DELETE route uses the URL as a path param (`<path:url>`) which can double-encode weird. body or query string is cleaner
> * the reorder endpoint doesn't validate the URLs actually exist (silent 0-row update)
> * `escapeHtml` is called on the URL in settings.js — just confirm that helper exists in the codebase, otherwise xss vector through a malicious instance URL
> 
> ping me once the auth + tests are in and i'll merge it. really good work on the actual fix

All done!

PR #400, opened 2026-04-28
Extract wishlist core logic from web_server.py

kettui:
## Summary
- Move wishlist-specific logic out of `web_server.py` into a dedicated `core/wishlist/` package
- Keep `core/wishlist_service.py` as a compatibility shim while the new package becomes the main home for wishlist logic
- Update import-side and search-side callers to use the shared wishlist helpers
- Add a dedicated `tests/wishlist/` suite so the extracted logic is covered directly
- Fix download cleanup after wishlist-triggered downloads by skipping unconfigured download clients

## What changed
- Added wishlist submodules for:
  - payload shaping
  - source-priority resolution
  - processing and cleanup
  - route adapters
  - state, reporting, selection, presence, and classification helpers
- Reduced `web_server.py` to thin delegation for the wishlist flow
- Moved wishlist-specific tests out of import-oriented test files and into `tests/wishlist/`
- Added regression coverage for the download cleanup path in `tests/downloads/test_download_orchestrator.py`

## Behavior
- The refactor is intended to preserve existing behavior
- The main functional change in the branch is the download cleanup fix after wishlist runs
- Old import paths are still supported through shims during the transition

## Validation
- Added broad unit coverage for the new wishlist modules
- Added regression coverage for download cleanup
- Verified touched files with compile checks during the extraction work


tests/downloads/test_downloads_post_processing.py
```diff
@@ -339,6 +339,7 @@ def test_fuzzy_context_matching_when_exact_key_missing():
     deps, rec = _build_deps(
         find_completed_file=lambda *a, **kw: (None, None),  # file not found
     )
+    monkeypatch.setattr(pp.time, 'sleep', lambda s: None)
```
kettui, line 342: Retry-induced `time.sleep(5)` calls were slowing down the tests

Nezreka:
Will check this out soon, thank you

PR #457, opened 2026-05-01
Clarify metadata source gating and Spotify auth flow

kettui:
### Spotify auth flow
- Streamline the Spotify auth experience so users automatically return to the app after completing authentication.
- Show only 1 button based on current auth status - "Authenticate" if not authenticated, "Disconnect" otherwise.
- Refresh the UI immediately when auth completes or when Spotify is disconnected.
- Remove descriptions and labels that refer to the legacy "Spotify-first" mental model, ie. no more talking about "fallback sources", or misleading iTunes references etc.

### Primary metadata source selection
- Show Spotify as locked in the primary metadata menu until an active session exists.
- Show Discogs as locked until a valid personal access token is configured.
- Redirect users to the relevant settings section when they try to select a locked source.

### Settings and status UI
- Keep the Spotify settings section, service status, and source selector aligned with the real auth state.
- Update the service indicators so they reflect the active metadata provider more clearly.
- Keep the UI copy consistent with the newer auth and provider model.

### Metadata helpers
- Centralize a few metadata-source label and disconnect helpers in the metadata layer.

image

### DIFF
--- a/webui/static/shared-helpers.js
+++ b/webui/static/shared-helpers.js
@@ -3145,6 +3145,10 @@ async function fetchAndUpdateServiceStatus() {
         // Cache for library status card
         _lastServiceStatus = data;
 
+        if (typeof syncSpotifySettingsAuthState === 'function') {
+            syncSpotifySettingsAuthState(data?.spotify || null);
+        }
+
         // Update service status indicators and text (dashboard)
         updateServiceStatus('spotify', data.spotify);
         updateServiceStatus('media-server', data.media_server);
--- a/webui/static/shared-helpers.js
+++ b/webui/static/shared-helpers.js
@@ -3194,10 +3194,15 @@ function updateServiceStatus(service, statusData) {
             currentMusicSourceName = sourceName;
         }
 
-        // Show/hide Spotify disconnect button based on connection state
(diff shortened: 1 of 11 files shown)


webui/static/shared-helpers.js
```diff
@@ -3160,45 +3196,133 @@ async function fetchAndUpdateServiceStatus() {
...
+        spotifyOption.title = spotifyAvailable
+            ? 'Spotify'
+            : 'Spotify authentication is required before this source can be selected.';
+    }
+
+    if (discogsOption) {
+        const discogsToken = document.getElementById('discogs-token');
+        const discogsAvailable = !!discogsToken?.value?.trim();
+        discogsOption.dataset.unavailable = discogsAvailable ? 'false' : 'true';
+        discogsOption.textContent = discogsAvailable ? 'Discogs' : '🔒 Discogs';
```
kettui, line 3220: Not sure what the status exactly is with discogs, but I inferred that having an access token is heavily encouraged, and as a result we probably want to prevent setting it as a primary metadata provider unless one is configured

Nezreka: I'd even encourage users to never use it XD

PR #467, opened 2026-05-02
Decouple metadata status from Spotify and remove redundant polling

kettui:
## Summary
- Split the dashboard/service-status data into:
  - `metadata_source` for the active primary metadata provider
  - `spotify` for Spotify-only auth and rate-limit state
- Moved status caching out of `web_server.py` into `core/metadata/status.py`
- Switched Spotify status updates to event-driven publishing instead of repeatedly probing from the status endpoints
- Reduced redundant watchlist and Spotify status polling across the UI
- Prevent service status card flickering at page load time by deferring the initial render to when the service status is actually retrieved. For example the first status card defaults to Spotify, which nowadays is rarely chosen by users, so it briefly shows the wrong value if we don't wait for the status query to return with the real values.

## What changed
- Watchlist page:
  - stopped the countdown timer from hammering `/api/watchlist/count` when no scan is scheduled
- Metadata status:
  - made the primary metadata source status generic instead of Spotify-shaped
  - kept Spotify-specific auth/rate-limit state separate
  - updated dashboard and settings consumers to use the split payload
- Backend cleanup:
  - centralized metadata status snapshot logic in `core/metadata/status.py`
  - added event-based Spotify status publishing from auth, disconnect, rate-limit, and config change paths
  - updated debug info and the rate-monitor feed to consume the shared cached snapshot

## Unrelated fixes
- Fixed the import pipeline so Spotify contexts keep real `spotify_*` IDs instead of falling back to numeric Deezer-like IDs, and added regressions for the context normalization and SoulSync library write path.

### DIFF
--- a/core/imports/context.py
+++ b/core/imports/context.py
@@ -32,6 +32,20 @@ def _first_id_value(*values: Any) -> str:
     return ""
 
 
+def _first_source_aware_id(source: str, *values: Any) -> str:
+    source_name = (source or "").strip().lower()
+    for value in values:
+        if value in (None, ""):
+            continue
+        text = str(value).strip()
+        if not text:
+            continue
+        if source_name.startswith("spotify") and text.isdigit():
+            continue
+        return text
+    return ""
+
+
 def extract_artist_name(artist: Any) -> str:
     if isinstance(artist, dict):
         return str(artist.get("name", "") or "")
@@ -209,6 +223,7 @@ def get_import_has_full_metadata(context: Optional[Dict[str, Any]]) -> bool:
 
 
 def get_import_source_ids(context: Optional[Dict[str, Any]]) -> Dict[str, str]:
+    source = get_import_source(context)
     track_info = get_import_track_info(context)
     original_search = get_import_original_search(context)
     search_result = get_import_search_result(context)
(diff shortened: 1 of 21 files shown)


Nezreka:
Nice, this looks good just needs a rebase

kettui:
> Nice, this looks good just needs a rebase

image

I'll do a rebase but it doesn't seem like it should be necessary

core/metadata/artwork.py
```diff
@@ -17,12 +19,193 @@
...
 
+def normalize_image_url(thumb_url: str | None) -> str | None:
+    """Convert media-server image URLs into browser-safe URLs."""
+    if not thumb_url:
+        return None
+
+    try:
+        if is_image_proxy_url(thumb_url):
+            # Already normalized for browser use; avoid wrapping it in another proxy layer.
+            return thumb_url
```
kettui, line 39: Snuck in this fix after rebasing the branch - I noticed that in some cases proxy image urls were already stored in the db, and certain features would end up trying to normalize those for the second time. This early return should prevent those

Nezreka:
> I'll do a rebase but it doesn't seem like it should be necessary

Yeah I think I just don't understand how rebase works lol. I'm scared the pr I merged before this will be lost if I merge this one without it being aware of the other pr. it doesn't work like that?

core/imports/context.py
```diff
@@ -32,6 +32,20 @@ def _first_id_value(*values: Any) -> str:
...
 
+def _first_source_aware_id(source: str, *values: Any) -> str:
+    source_name = (source or "").strip().lower()
+    for value in values:
+        if value in (None, ""):
+            continue
+        text = str(value).strip()
+        if not text:
+            continue
+        if source_name.startswith("spotify") and text.isdigit():
```
kettui, line 43: Noticed on my local dev setup (which uses the SS standalone server) that `get_import_source_ids` would end up storing deezer id's in spotify-specific fields. This means that it's quite likely that standalone server users could have some poor data in their db

kettui:
> I'm scared the pr I merged before this will be lost if I merge this one without it being aware of the other pr. it doesn't work like that?

As long as the changes made in the target branch don't conflict with the PR (which is indicated in the github UI), merging the PR is safe. When a PR is merged, it doesn't overwrite the whole file that was touched, it only merges the parts that were changed in the PR branch

PR #495, opened 2026-05-04
Refactor/download source plugins

Nezreka:
# Multi-source Download Refactor 

unifies the multi-source download dispatcher behind a real plugin contract + central engine. follow-up to the metadata layer Cin's been working on (POC #498), same architectural smell.

layout

### Plugin contract + registry (Cin-1)
- new `core/download_plugins/` package: `DownloadSourcePlugin` Protocol, `DownloadPluginRegistry`
- every download client (soulseek / youtube / tidal / qobuz / hifi / deezer / lidarr / soundcloud) explicitly inherits `DownloadSourcePlugin` so it's obvious they implement the contract
- shared dataclasses (`TrackResult`, `AlbumResult`, `DownloadStatus`, `SearchResult`) lifted out of `core/soulseek_client.py` into `core/download_plugins/types.py` — plugins no longer import from a sibling source just to satisfy the contract

### Central engine + worker
- `core/download_engine/`: state, threading, search retry, rate-limits, fallback dispatch
- `BackgroundDownloadWorker.dispatch(source, target_id, impl_callable, ...)` lifts ~490 LOC of duplicated thread-spawn boilerplate out of every streaming source
- per-source `RateLimitPolicy` declared on the plugin, engine reads at register time
- nested-dict state layout — per-source iteration is O(source_records), not O(total_records)
- per-source RLock sharding — Deezer progress callbacks don't block Tidal writes
- atomic `update_record_unless_state(skip_if_state_in=('Cancelled',))` so terminal writes preserve a user's mid-flight cancel

### Generic orchestrator API (Cin-2 / Cin-5)
- `orch.client(name)` — alias-aware lookup (legacy `deezer_dl` resolves to `deezer`)
- `orch.configured_clients()` — returns `{name: client}` for every initialized + configured source
- `orch.reload_instances(name=None)` — generic reload dispatch
- singleton: `get_download_orchestrator()` / `set_download_orchestrator()` matching the `get_metadata_engine()` shape
- `web_server.py` boots via `set_download_orchestrator(...)` so factory + global share state

### `soulseek_client` global → `download_orchestrator` (Cin-6)
- the global handle has been a `DownloadOrchestrator` for ages, the legacy name was misleading
- renamed everywhere: web_server.py, api/, core/downloads/*, core/search/*, core/streaming/*, services/sync_service.py
- module path `core.soulseek_client` and class `SoulseekClient` (the actual soulseek-only client) unchanged

### Bug fixes from Cin's quick-findings review
- 🔴 `engine.cancel_download(source_hint='deezer_dl')` was falling through to soulseek because the alias never made it to the engine's plugin map → added alias map on the engine
- 🔴 `_resolve_source_chain` filtered hybrid_order against canonical names only → legacy `deezer_dl` config silently dropped deezer; now normalizes through `registry.get_spec()`
- 🟡 worker's terminal write was a read-then-write split that could clobber a Cancelled state → atomic guard

### Bug fixes from JohnBaumb's review
- engine state was a single dict with composite keys → switched to nested-dict so per-source iteration is O(source_records)
- single `state_lock` serialized progress callbacks across every source → per-source RLock sharding
- download clients returned None on engine-not-wired → silent drop with no user feedback; now raise so `download_with_fallback` surfaces a real warning + tries the next source
- `engine.get_all_downloads()` had no exclude param → monitor was double-fetching slskd transfers (already pulled via the transfers/downloads endpoint earlier in the same loop); added `exclude=()` param

## ⚠️ BREAKING — heads up for in-flight PRs

The per-source attribute aliases on the orchestrator are GONE.

```python
# Before (works on main):
orch.soulseek
orch.youtube
orch.tidal
orch.qobuz
orch.hifi
orch.deezer_dl
orch.lidarr
orch.soundcloud

# After (this PR):
orch.client('soulseek')
orch.client('youtube')
orch.client('tidal')
# etc — alias-aware, so 'deezer_dl' still resolves
```

Anything that reached `orch.<source>` directly will hit AttributeError. Migrate to `orch.client('<source>')`. Earlier draft of the PR description claimed the aliases were preserved — they aren't, that was stale (sorry).

The global handle was also renamed:

```python
# Before:
soulseek_client.something(...)

# After:
download_orchestrator.something(...)
```

If you have a PR open against this subsystem, both renames will hit you.

## Test plan

- [x] `python -m pytest` — full suite green (2074 pass, ~100 new tests across engine + worker + orchestrator + plugin conformance + SoundCloud preview filter)
- [x] download from each configured source (soulseek / youtube / tidal / qobuz / hifi / deezer / lidarr / soundcloud) — file lands, status flips Completed, post-process runs
- [x] hybrid mode search across multiple sources still walks the chain
- [x] cancel mid-flight stays Cancelled (doesn't get clobbered to Errored / Completed)
- [x] settings reload picks up new download path on every source
- [x] hifi instance add/remove/toggle still works
- [x] deezer ARL endpoints (status / playlists / playlist tracks) work
- [x] track redownload modal lists candidates from every configured source


kettui:
Some quick findings:

core/download_engine/engine.py:L233: 🔴 bug: source_hint='deezer_dl' never matches _plugins, so Deezer cancel requests fall through to Soulseek and no-op. Resolve aliases through the registry before routing.

core/download_orchestrator.py:L203: 🔴 bug: _resolve_source_chain() filters against canonical names only, so legacy deezer_dl values from hybrid_order / hybrid_primary / hybrid_secondary get dropped and Deezer disappears from hybrid mode. Normalize names through registry.get_spec() and keep the fallback order deterministic.

core/download_engine/worker.py:L258: 🟡 risk: terminal writes are not atomic with the cancel check, so a cancel that lands between the snapshot and the update can be overwritten back to Completed/Errored. Hold the engine lock across the check+write or add an atomic guard; _mark_terminal() has the same split.

core/download_engine/engine.py
```diff
@@ -0,0 +1,417 @@
...
+            return dict(record) if record is not None else None
+
+    def iter_records_for_source(self, source_name: str) -> Iterator[DownloadRecord]:
+        """Yield SHALLOW COPIES of every record owned by a source.
+        Holds the lock briefly to snapshot, then yields outside the
+        lock so callers can spend arbitrary time on each record."""
+        with self.state_lock:
+            snapshot = [
+                dict(record)
+                for (source, _), record in self._records.items()
```
JohnBaumb: O(total_records) not O(source_records)

This iterates every record across all sources just to filter one source.

Nezreka: Great catch! fixing now

core/download_engine/engine.py
```diff
@@ -0,0 +1,429 @@
...
+    Thread safety: every state mutation goes through ``state_lock``.
+    Read-only accessors (``get_record``, ``iter_records_for_source``)
+    take the lock briefly and return a SHALLOW COPY so the caller
+    can iterate without holding the lock. Callers that need to
+    mutate a record should use ``update_record`` which takes the
+    lock and applies the patch atomically.
+    """
+
+    def __init__(self) -> None:
+        self.state_lock = threading.RLock()
```
JohnBaumb: The single state_lock serializes progress updates across all sources. 

Before, each source had its own download lock, so Deezer/YouTube/Tidal workers never blocked each other. 

Under concurrent multi-source downloads, the worker threads will now fight onthis one RLock for every progress callback, may causes slowdowns and waiting.

not sure if this is on purpose, but I think it can be better so they don't block each other during progress calls, even though the delay time will be in the microseconds XD

maybe do per-source lock sharding?

JohnBaumb: it's also more of a nit-pick, because the delay time will be so short.

Nezreka: got it :D

core/downloads/monitor.py
```diff
@@ -266,31 +267,28 @@ def _get_live_transfers(self):
...
-                                   soulseek_client.hifi, soulseek_client.deezer_dl, soulseek_client.lidarr,
-                                   getattr(soulseek_client, 'soundcloud', None)]:
-                    if _dl_client:
-                        try:
-                            all_downloads.extend(run_async(_dl_client.get_all_downloads()))
-                        except Exception:
-                            pass
+                if download_orchestrator and hasattr(download_orchestrator, 'engine'):
+                    try:
+                        all_downloads = run_async(download_orchestrator.engine.get_all_downloads())
```
JohnBaumb: This doesn't exclude soulseek so it is double getting, previously it would exclude soulseek here. repeating what it does on line 259

core/hifi_client.py
```diff
@@ -571,71 +575,31 @@ def _to_track_result(self, track: Dict, quality_info: Dict) -> TrackResult:
...
-            return download_id
-
-        except Exception as e:
-            logger.error(f"Failed to start HiFi download: {e}")
+        if '||' not in filename:
+            logger.error(f"Invalid filename format: {filename}")
+            return None
+        if self._engine is None:
+            logger.error("HiFi client has no engine reference — cannot dispatch download")
             return None
```
JohnBaumb: All the download plugins return none here, so no user feedback, the download disappears.

core/download_engine/engine.py
```diff
@@ -0,0 +1,451 @@
...
+        ``engine.worker`` / read state / etc. Plugins that haven't
+        been migrated to the engine yet simply don't define
+        ``set_engine`` — they keep their pre-engine behavior
+        unchanged.
+
+        Also reads the plugin's declared ``RateLimitPolicy`` (via
+        the ``rate_limit_policy()`` method or ``RATE_LIMIT_POLICY``
+        class attribute) and applies it to the worker. Plugins that
+        don't declare a policy get the conservative default
+        (concurrency=1, delay=0).
```
JohnBaumb, line 120: the plugin can register here, but if it's none it stays registered. downloads silently fail, no user feedback except in logs.

JohnBaumb:
PR description says orchestrator.soulseek / .youtube / etc. attribute access is still preserved, but the self.soulseek = self._safe_init(...) assignments are removed with no getattr fallback. If they were removed on purpose then the pr description is stale, if not on purpose, at least you know.

This is fine, pending pr's affected by this will just need to switch to the new methods, but they will probably be surprised by getattr errors lol

JohnBaumb:
Suggested test additions:

register_plugin when set_engine() raises -- There's no test that verifies the system behaves correctly when a plugin's set_engine throws. The test_download_raises_when_engine_not_wired tests exist (your feedback was already incorporated!) but no test covers the path where registration succeeds yet _engine stays None because set_engine failed. You'd want to verify downloads surface an error to the user rather than vanishing.

engine.get_all_downloads() excludes sources -- There's a test_engine_get_all_downloads_skips_excluded_sources but no test covering the monitor's actual usage pattern (calling engine.get_all_downloads() after already fetching Soulseek separately). Not critical since that's more of a behavior bug than a unit test gap.

cancel_download during active _download_sync -- The worker tests cover cancelled state preservation after impl returns, but there's no test for a cancel arriving while _download_sync is mid-execution (testing _is_cancelled polling inside the download loop). The pinning tests check final state, not the mid-flight interruption.

configured_clients() when is_configured() raises -- The method has a try/except but no test verifies that a broken is_configured() doesn't crash the iteration.

Rate limit delay with concurrent sources -- test_per_source_locks_dont_block_each_other exists for the state lock, but there's no test proving YouTube's 3s delay doesn't block Tidal's downloads (the worker semaphores + delays are per-source but share the engine).

PR #496, opened 2026-05-05
fix: skip recursive chown on startup when UIDs are already correct

dlynas:
## Summary

- The unconditional `chown -R` on every container start walks the entire `/app` tree on each `docker compose up`, including large music libraries with thousands of files — causing a significant delay before the app starts
- The recursive walk is now skipped when UIDs/GIDs are already correct; only the directory nodes themselves are chowned at startup
- The full recursive `chown -R` still runs inside the UID-change branch where it is actually needed (lines 38–42)

## Test plan

- [x] Start container with correct PUID/PGID — startup should be instant, no permission walk
- [x] Start container with mismatched PUID/PGID — recursive chown should still run as before

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -35,9 +35,15 @@ if [ "$CURRENT_UID" != "$PUID" ] || [ "$CURRENT_GID" != "$PGID" ]; then
         usermod -o -u "$PUID" soulsync
     fi
 
-    # Fix ownership of app directories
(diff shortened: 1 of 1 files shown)


Nezreka:
This seems great! Give it a test and mark off your test plan if all is good when you have time

dlynas:
Thanks!!
This PR is already up and running in production for me and works as intended!

PR #433, opened 2026-04-30
feat: full discography, explicit badges, smart selection, and sorting

dlynas:
Pagination: raise discography limit from 50 to 500 across all three API call sites (artist-detail page, discography download modal, source-detail view). Deezer/iTunes/Spotify already paginated internally; Discogs updated to paginate all pages instead of a single capped request.

Explicit flag: add explicit field to the Album dataclass in every metadata source (Deezer, Spotify, iTunes, Discogs, MusicBrainz, Hydrabase). Deezer extracts explicit_lyrics, iTunes extracts collectionExplicitness, Hydrabase forwards the field from the server response; Discogs/MusicBrainz/Spotify stay None. The field threads through _build_discography_release_dict and _build_artist_detail_release_card so it reaches the frontend on every release object. A small "E" badge renders next to explicit album titles in both the discography download modal and artist-detail page cards.

Smart selection: "Select Best" button in the discography download modal groups visible cards by normalised title (edition/variant suffixes stripped) and within each group checks only the highest-scoring version. Scoring
mirrors the existing Python _variant_score logic: explicit > non-clean >
no variant suffix > more tracks > newer release date.

Sort controls: Date desc/asc and Name A-Z/Z-A dropdown added to the discography modal filter bar. Sorting reorders DOM nodes in place so checked state is preserved across sort changes.

UI: discography card titles now wrap to full text instead of truncating with ellipsis; card artwork and checkbox pin to the top when titles wrap.

PR #527, opened 2026-05-08
feat: raise discography limit to 500 and paginate Discogs artist releases

dlynas:
## Summary

- Raises the discography album fetch limit from 50 → 500 across all three call sites: artist-detail page, discography download modal, and source-detail view
- Deezer, iTunes, and Spotify already paginate internally — no changes needed there
- Discogs is updated to paginate through **all pages** (100 results/request, the API max) before filtering, so the 500-album limit applies to qualified albums rather than raw API rows

## Changes

- `core/discogs_client.py` — `get_artist_albums()` now loops through all Discogs pages, collecting items before the dedup/filter pass; early-exit guard handles `limit=0` (unlimited) correctly
- `web_server.py` — both the artist-detail and discography-modal endpoints: `limit=50` → `limit=500`
- `core/artist_source_detail.py` — same lift for the source-detail view

## Notes

This is part of a refactor of my previous PR (#433, now set to draft) — splitting the original combined branch into focused, individually reviewable PRs. This is the first of those.

## Testing

Tested in situ on my own server via a locally-built Docker image — confirmed artists with large catalogs now return well beyond the previous 50-release cap.

### DIFF
--- a/core/artist_source_detail.py
+++ b/core/artist_source_detail.py
@@ -126,7 +126,7 @@ def build_source_only_artist_detail(
             allow_fallback=True,
             skip_cache=False,
             max_pages=0,
-            limit=50,
+            limit=500,
             artist_source_ids={source: artist_id},
             dedup_variants=False,
         ),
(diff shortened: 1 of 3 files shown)


dlynas:
Closing — upstream implemented their own pagination in 2.5.3.

PR #590, opened 2026-05-14
Migrate stats page to React

kettui:
## Summary

- migrate `stats` from the legacy shell to the React route host
  - basically a 1:1 port, no functional changes that I'm aware of
  - selected time period is persisted to the url
- replace the global `Chart.js` dependency with route-local `Recharts`
- add a typed stats route slice with route-local queries, helpers, charts, and UI state
- route legacy playback and artist-detail handoffs through the explicit shell bridge
- remove the old stats HTML, JS, and CSS implementation after cutover
- update / standardize migration docs a bit
  - helps keeping the migration progress transparent, as well as providing useful context for future migration work


Nezreka:
noice, couple things before I merge:

`8c6d803d` ran prettier across `core.js` and `shared-helpers.js` — ~7235 lines of indent/comma churn for one real fix (`if (btn.id === 'stats-sync-btn') return;` in core.js). buries it.

also `navigateToArtistDetail` got declared twice in two places — `webui/static/shell-bridge.js:171` + `:180` (object literal, JS keeps the last one) and `webui/src/platform/shell/globals.d.ts:38` + `:47` (TS overload-merges, no error, but the two `options` shapes diverge). looks like a rebase leftover.

other than that, it's gold.

PR #594, opened 2026-05-14
Add Telegram thread

Skowll:
## Summary
This pull request adds support for a new `thread_id` field in the Telegram notification automation.
With this change, notifications can be sent directly to a specific Telegram thread/topic instead of always being posted in the main chat.

## Motivation
Some Telegram group setups rely on topics to organize discussions and automated messages.
Adding `thread_id` support makes the notification system more flexible and allows better message routing inside Telegram forums.

## Changes
- Added a new `thread_id` field to the Telegram notification configuration.
- Updated the Telegram notification sending logic to forward this value to the Telegram Bot API.
- Enabled notifications to target a specific thread/topic when `thread_id` is provided.
- Preserved the existing behavior for configurations that do not define a `thread_id`.

## Testing

- Verified that messages are still sent correctly when no `thread_id` is configured.
- Verified that messages are delivered to the expected Telegram topic when `thread_id` is set.
- Confirmed that the change does not affect existing Telegram notification configurations.

## Notes
This is my first contribution to the project, so I would appreciate an especially careful review before merging.
I have tested the change locally and the feature behaves as expected, but please review it with appropriate caution.


Nezreka:
hey skowll, thanks for the pr. one blocker:

**🔴 backward compat break** — `core/automation_engine.py:1011` always includes `message_thread_id` in the telegram payload, even when the user hasn't set a thread (thread_id is `""`). telegram api wants integer, not empty string — risks `Bad Request: invalid integer` for every existing user who hasn't touched the new field.

fix:
```python
payload = {"chat_id": chat_id, "text": message, "parse_mode": "HTML"}
if thread_id:
    try:
        payload["message_thread_id"] = int(thread_id)
    except ValueError:
        pass  # invalid — fall back to main chat
resp = requests.post(..., json=payload, timeout=10)
```

small cleanups while you're at it:

- `stats-automations.js` — `threadID` should be `threadId` (camelCase matches the surrounding `botToken` / `chatId`)
- `stats-automations.js` — double space in `'thread_id':  (overlay...)`
- `wishlist-tools.js` help — `Thread ID: Thread ID` doesn't explain anything. try `Optional — only set if your group uses Telegram topics. Leave blank for the main chat.`

Skowll:
Hello,
I think I've fixed all the points you raised. Feel free to check it over again and let me know if anything else is needed!

PR #644, opened 2026-05-19
Centralize artist-detail page hand-off logic, make it URL-driven

kettui:
Followup improvements related to https://github.com/Nezreka/SoulSync/pull/636

## Summary
- Add `/artist-detail/:source/:id` as the canonical artist-detail route in TanStack Router
- Keep the legacy artist-detail page as the renderer behind a thin route handoff
- Convert user-triggered artist-detail navigation to semantic links so clicks stay SPA-driven
- Remove custom legacy routing/origin-stack behavior in favor of browser history and router state
- Cancel stale similar-artists loading when the artist-detail route changes or unmounts
  - This was an unrelated improvement, but I noticed it was preventing another artist detail page from loading if the similar artist load was still running in the background, at least in dev mode

## Details
- Added route and bridge typings for the artist-detail handoff
- Replaced imperative artist-detail navigation across search, library, discover, downloads, stats, sync, API monitor, media player, and shared card helpers
- Removed dead helper wrappers and page-specific bridge plumbing
- Tightened route cleanup so old similar-artists requests stop when the user navigates away

### DIFF
--- a/webui/src/routes/artist-detail/$source/$id.tsx
+++ b/webui/src/routes/artist-detail/$source/$id.tsx
@@ -18,7 +18,6 @@ function ArtistDetailPage() {
 
     const normalizedSource = source.toLowerCase() === 'library' ? null : source.toLowerCase();
     bridge.navigateToArtistDetail(id, '', normalizedSource, {
-      skipOriginPush: true,
       skipRouteChange: true,
     });
   }, [bridge, id, source]);
--- a/webui/src/routes/artist-detail/$source/$id.tsx
+++ b/webui/src/routes/artist-detail/$source/$id.tsx
@@ -1,0 +1,27 @@
+import { createFileRoute } from '@tanstack/react-router';
+import { useLayoutEffect } from 'react';
+
+import { useShellBridge } from '@/platform/shell/route-controllers';
+
+export const Route = createFileRoute('/artist-detail/$source/$id')({
+  component: ArtistDetailPage,
+});
+
+// Thin legacy handoff: TanStack owns the URL shape here, but the vanilla JS
+// artist-detail page still renders the actual experience for now.
+function ArtistDetailPage() {
+  const bridge = useShellBridge();
+  const { source, id } = Route.useParams();
+
+  useLayoutEffect(() => {
+    if (!bridge) return;
+
+    const normalizedSource = source.toLowerCase() === 'library' ? null : source.toLowerCase();
+    bridge.navigateToArtistDetail(id, '', normalizedSource, {
+      skipOriginPush: true,
+      skipRouteChange: true,
+    });
+  }, [bridge, id, source]);
+
+  return null;
+}
--- a/webui/src/routes/artist-detail/$source/$id.tsx
+++ b/webui/src/routes/artist-detail/$source/$id.tsx
@@ -1,5 +1,5 @@
 import { createFileRoute } from '@tanstack/react-router';
-import { useLayoutEffect } from 'react';
+import { useEffect, useLayoutEffect } from 'react';
 
 import { useShellBridge } from '@/platform/shell/route-controllers';
 
@@ -8,7 +8,8 @@ export const Route = createFileRoute('/artist-detail/$source/$id')({
 });
 
 // Thin legacy handoff: TanStack owns the URL shape here, but the vanilla JS
-// artist-detail page still renders the actual experience for now.
+// artist-detail page still renders the actual experience for now. The route
+// owns cancellation so similar-artist loading stops when this page changes.
 function ArtistDetailPage() {
   const bridge = useShellBridge();
   const { source, id } = Route.useParams();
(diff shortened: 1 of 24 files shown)


webui/src/routes/artist-detail/$source/$id.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const Route = createFileRoute('/artist-detail/$source/$id')({
+  component: ArtistDetailPage,
+});
+
+// Thin legacy handoff: TanStack owns the URL shape here, but the vanilla JS
+// artist-detail page still renders the actual experience for now. The route
+// owns cancellation so similar-artist loading stops when this page changes.
+function ArtistDetailPage() {
+  const bridge = useShellBridge();
+  const { source, id } = Route.useParams();
```
kettui, line 15: tanstack router automatically parses these based on the route definition above (`/artist-detail/$source/$id`)

webui/src/routes/artist-detail/$source/$id.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  useLayoutEffect(() => {
+    if (!bridge) return;
+
+    const normalizedSource = source.toLowerCase() === 'library' ? null : source.toLowerCase();
+    bridge.navigateToArtistDetail(id, '', normalizedSource, {
+      skipRouteChange: true,
+    });
+
+    return () => {
```
kettui, line 25: This callback is called whenever the user navigates from one artist-detail page to another, or to an entirely different page

webui/static/api-monitor.js
```diff
@@ -2398,7 +2417,7 @@ async function openWatchlistArtistDetailView(artistId, artistName) {
                 </div>
 
                 <div class="watchlist-detail-actions">
-                    <button class="watchlist-detail-discog-btn watchlist-detail-discog-action">View Discography</button>
+                    <a class="watchlist-detail-discog-btn watchlist-detail-discog-action" href="${discogHref}" ${discogId ? 'onclick="closeWatchlistArtistDetailView()"' : 'aria-disabled="true" tabindex="-1" style="pointer-events:none;opacity:0.5;text-decoration:none;color:inherit;"'}>View Discography</a>
```
kettui, line 2420: All links / buttons / etc. to the artist-detail page are now actual link elements in html, which means that the artist-detail page rendering is driven entirely by the new React route that we added here

webui/static/library.js
```diff
@@ -951,27 +869,12 @@ function initializeArtistDetailPage() {
                 artistDetailPageState.completionController = null;
             }
 
-            const stack = artistDetailPageState.originStack || [];
```
kettui, line 954: no need for manual originStack logic anymore since we're using native browser navigation

webui/static/shell-bridge.js
```diff
@@ -185,10 +165,58 @@ window.SoulSyncWebShellBridge = {
...
+        artistSource: source,
+    });
+}
+
+function _isModifiedLinkClick(event) {
+    return event.metaKey || event.ctrlKey || event.shiftKey || event.altKey;
+}
+
 window.addEventListener('popstate', syncActivePageFromLocation);
+document.addEventListener('click', _handleShellLinkClick, true);
```
kettui, line 221: this click interceptor allows us to use native `<a>` link elements on the legacy webui side, while still getting a "SPA-like" navigation UX when clicking them. By default native links would reload the page, but this let's us delegate any relevant navigations to Tanstack Router, which avoids reloading the page (and thus gives a smoother user experience)

webui/static/library.js
```diff
@@ -951,27 +869,12 @@ function initializeArtistDetailPage() {
...
-                    navigateToArtistDetail(target.id, target.name, target.source, { skipOriginPush: true });
-                    return;
-                }
-                // target.type === 'page' — fully exit the artist-detail chain
-                artistDetailPageState.currentArtistId = null;
-                artistDetailPageState.currentArtistName = null;
-                artistDetailPageState.originStack = [];
-                navigateToPage(target.pageId);
+            if (window.history.length > 1) {
+                window.history.back();
```
kettui, line 873: the "Back" button now simply delegates to native history api for navigation

webui/static/shared-helpers.js
```diff
@@ -3648,7 +3666,7 @@ async function loadSimilarArtists(artistName) {
                             `;
                         } else {
                             // Lazy load images for similar artists that don't have them
-                            lazyLoadSimilarArtistImages(container);
+                            await lazyLoadSimilarArtistImages(container, signal);
```
kettui, line 3669: these similar artist loads would previously just keep happening in the background even after navigating away from the artist-detail page

webui/static/api-monitor.js
kettui: Given that they are now proper links, one can also do useful things like open the link in a new tab. Most (if not all) internal links inside SS are not url-driven like this, so they don't get the same benefits yet. Migrating pages over to React (either completely or partially like in this PR) will naturally allow us to make navigation driven by the url, and over time more and more pages can be directly linked to / bookmarked / opened in a new tab / refreshed, etc.

PR #686, opened 2026-05-24
feat(webui): migrate import page to React

kettui:
## Summary
- Move the `/import` experience into the React shell while keeping the existing workflow intact.
- Preserve album matching, singles matching, auto-import review, and the processing queue.
- Keep `/import` landing on the Albums tab, with URL state for tab selection and auto-import filters.

## Key Changes
- Add React route slices for `/import`, `/import/album`, `/import/singles`, and `/import/auto`.
- Split import logic into route-local API, helper, store, and UI modules.
- Expand shared form primitives for the new import controls.
- Remove legacy import DOM and activation code from `webui/index.html` and `webui/static/init.js`.
- Trim the old global import JS/CSS and move route styling into `import-page.module.css`.
- Update the shell route manifest, generated route tree, migration docs, and smoke coverage.
- Remove obsolete legacy import frontend pytest coverage.

## Coverage
- Added or updated unit tests for import API, route behavior, and form components.
- Updated shell route smoke tests for the new React-owned import flow.


kettui:
Noteworthy that while there has been some component extraction (especially in this PR) for a bunch of common / shared components, there's still quite a few different things that could be extracted, now that more pages are slowly being moved over. I don't want to bloat this PR further so I won't include that work in this PR, as it would touch other pages quite heavily as well

PR #731, opened 2026-05-29
Fix album-bundle release picker grabbing the wrong album — gate by title relevance (#730)

### LINKED ISSUE #730: Usenet/Torrent album download grabs the wrong release — picker ignores album title
**What happens:** Requesting an album download via a Prowlarr-backed source (Usenet or Torrent) can download a *different* album. Example: requesting David Bowie – *"Heroes" (2017 Remaster)* downloaded *Scary Monsters and Super Creeps* instead.

**Root cause:** `pick_best_album_release` (`core/download_plugins/album_bundle.py`) ranks candidates only by seeders/grabs → quality → size, with **no relevance check against the requested album title**. Since Prowlarr returns broad fuzzy matches (a "Heroes" search also returns other Bowie releases), the most-*popular* result wins regardless of whether it's the right album — and *Scary Monsters* had roughly 16× the grabs.

**Scope:** Affects **Usenet and Torrent** (both go through `pick_best_album_release`). Soulseek is unaffected — its album bundle uses the title/artist/track-coverage-aware pre-flight (`_score_album_folder`, gated by `_ALBUM_PREFLIGHT_MIN_SCORE`), which already validates the folder matches the requested album.

**Proposed fix:** Gate candidates by album-title relevance before ranking, and refuse (fall back to the per-track flow) rather than download a mismatch when nothing qualifies.

tyler-rich:
Closes #730.

## Problem

`pick_best_album_release` (`core/download_plugins/album_bundle.py`) ranked Prowlarr candidates purely by popularity (seeders/grabs) → quality → size, with **no check that the release is actually the requested album**. Prowlarr returns broad fuzzy matches, so the most-*popular* result could win regardless — a David Bowie *"Heroes" (2017 Remaster)* request downloaded *Scary Monsters and Super Creeps* instead (~16× the grabs).

Affects **Usenet and Torrent** (both go through this shared picker). Soulseek is unaffected — it uses the title/artist/track-coverage-aware album pre-flight (`_score_album_folder`, gated by `_ALBUM_PREFLIGHT_MIN_SCORE`) rather than this picker.

## Fix

- New `album_title_relevance()` — token-coverage of the album's *core* title (year/edition/remaster noise stripped) against the release name; robust to the codec/year/group noise that pads release names.
- `pick_best_album_release` now gates candidates by relevance (floor `0.55`) **before** the popularity ranking, and **refuses** (returns `None` → caller falls back to per-track search) rather than download a mismatch when nothing qualifies.
- Both `usenet.py` and `torrent.py` call sites pass `album_name`/`artist_name`; callers that omit it keep the legacy popularity ranking (back-compat).

## Testing

- Reproduces the exact Heroes-vs-Scary-Monsters case (grabs 960 vs 15698, album `"Heroes" (2017 Remaster)`) and asserts Heroes wins — plus relevance scoring, no-match refusal, multi-token coverage, and no-album-name back-compat.
- Verified locally (`python web_server.py`): the correct album is now selected.
- `ruff` + `compileall` + `pytest` green.

Based on `dev`; independent of #728.

### DIFF
--- a/core/download_plugins/album_bundle.py
+++ b/core/download_plugins/album_bundle.py
@@ -21,6 +21,7 @@ folder scan.
 
 from __future__ import annotations
 
+import re
 import shutil
 import time
 import uuid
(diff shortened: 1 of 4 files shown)


Nezreka:
hey, thanks for this — really appreciate you chasing the album-bundle stuff down, and the Heroes/Scary Monsters writeup made it super easy to follow. tests are clean and i can confirm it fixes the exact case you described.

couple things before i pull it in:

**rebase needed (trivial).** i just merged #728 to dev, so this needs a quick rebase. good news: the actual source files (`album_bundle.py`, `torrent.py`, `usenet.py`) all auto-merge clean — the *only* conflict is `tests/test_album_bundle.py` where we both added to the same import block + appended test funcs. like a 2 minute fix.

**the one real thing — `artist_name` is passed but never used.** both call sites hand it in, but `pick_best_album_release` / `album_title_relevance` only ever look at the album title. so there's no artist check at all, and for single-word album names that leaves the same hole #730 was about. i poked at the scoring directly:

| request | release | score |
|---|---|---|
| `Heroes` | **Willie Nelson** – Heroes | 1.00 |
| `Heroes` | Bowie – Heroes (Live Berlin) | 1.00 |
| `Thriller` | VA – Greatest **Thriller** Hits | 1.00 |

so a wrong-artist / live / comp release with the right word in the title still passes the gate, and then popularity picks the winner — which is basically the original bug again for common one-word titles. since you've *already* plumbed `artist_name` through both callers, folding artist token-coverage into the gate would close that with almost no extra work. that's the one i'd want done before merge.

**minor (fine as a follow-up):** symbol/short titles trip the noise fallback. e.g. `÷ (Deluxe)` strips down to nothing, falls back to raw tokens, and ends up with `deluxe` as the "core" token — so it both refuses the correct release AND matches random "deluxe" releases. low frequency, don't sweat it for this PR, just noting it.

the refuse-and-fall-back-to-per-track behavior when nothing matches is exactly right btw, like that a lot. nothing here is a dealbreaker — it's a clear improvement and it degrades safely. just want the artist gating in since it's the actual heart of #730 and it's basically free at this point.

thanks again 🙏

PR #776, opened 2026-06-02
Fix false library match between normal and Extended Mix titles

kekkokk:
## Summary
- Fixes Download Missing marking distinct playlist tracks (e.g. normal vs Extended Mix) as already in library at 100% match.
- Featuring cleanup no longer strips version suffixes after bracket/dash normalization.
- Adds regression tests for `Latinamerica (feat. Vika)` vs Extended Mix.
## Test plan
- [ ] `pytest tests/test_track_title_version_distinction.py`
- [ ] Spotify playlist with normal + Extended Mix: only the missing version is queued after Begin Analysis

### DIFF
--- a/database/music_database.py
+++ b/database/music_database.py
@@ -7296,6 +7296,22 @@ class MusicDatabase:
         cleaned = re.sub(r'\s*\(with\s+[^)]*\)', '', cleaned, flags=re.IGNORECASE)
         cleaned = re.sub(r'\s*\[with\s+[^\]]*\]', '', cleaned, flags=re.IGNORECASE)
 
+        # Featuring in parentheses/brackets — must run BEFORE bracket flattening.
+        # Otherwise ``feat.`` removal after flattening is greedy and strips
+        # version suffixes ("Extended Mix", "Radio Edit", …) from the same title.
+        cleaned = re.sub(
+            r'\s*\([^)]*(?:feat\.?|ft\.?|featuring)[^)]*\)',
+            '',
+            cleaned,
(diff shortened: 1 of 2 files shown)


Nezreka:
hey, thanks for this — the core fix is solid, normal vs extended mix not collapsing anymore is exactly right and the tests cover it well.

one thing before this goes in though. the new parenthesized feat/ft regex doesn't have word boundaries, so `ft` matches *inside* words. anything in parens that just happens to contain "ft" or "feat" as a substring gets stripped whole. ran it on the branch:

- `Get Lucky (Daft Punk Remix)` -> `get lucky`  (the "ft" in **Da**ft)
- `Harder Better (Soft Mix)` -> `harder better`  (**So**ft)
- `One More Time (Aftermath Edit)` -> `one more time`  (A**ft**ermath)

old code kept all of those fine (`get lucky daft punk remix` etc), so it's a regression. and it's kinda the same bug this PR is killing, just flipped — a daft punk remix now collapses to the original title, so if i only own "Get Lucky" the remix shows as already in library. daft punk remixes aren't exactly rare lol.

the unparenthesized half is fine btw, that one's bounded with `\s+...\s+`. it's just the bracketed regexes that need it.

fix is small — word-boundary the tokens, something like `\b(?:feat|ft|featuring)\b\.?`. `\bft\b` won't match inside "daft"/"soft" but still catches real `(feat. X)`. if you can throw a `Daft Punk Remix` / `Soft Mix` case into the regression test too that'd be perfect.

other than that it's good to go 👍

PR #779, opened 2026-06-03
Feature/spotify public full playlist

Nezreka:
# Full public playlists on the "Spotify link" tab (no credentials)

## what this fixes

If you paste a Spotify playlist link in **Sync → Spotify link**, you used to
only get the first ~100 tracks. That's because the no-login path scrapes
Spotify's **embed widget**, and the embed only ever contains ~100 songs — it
can't page past that. So a 230-track playlist came in truncated.

This pulls the **full** list with no Spotify credentials.

## how

Getting the whole list no-login means talking to Spotify's private API the way
their web player does — which now requires client-auth headers Spotify
**rotates constantly**. I tried doing that in-house (grab the anonymous token
out of Spotify's own embed page, then paginate the API) — the token works, but
Spotify **429s** the request without the rest of the web player's rotating
client-auth. Chasing that by hand is a losing game (it's literally why spotDL
stopped doing it themselves).

So this leans on **SpotipyFree** — the maintained, no-credentials `spotipy`
drop-in that spotDL uses, which tracks all that rotating machinery for us. It
returns spotipy-shaped data, which we normalise into the same shape the embed
scraper already produces.

- `core/spotify_public_api.fetch_public_playlist_full` — builds a SpotipyFree
  client and pages the playlist (`playlist` for name/owner, `playlist_items` +
  `next` for tracks), normalising to the embed shape. The client is injectable
  so the logic is unit-testable without the library or the network.
- `core/spotify_public_scraper.fetch_spotify_public` — tries the full fetch for
  playlists; on **any** failure (library not installed, Spotify changed
  something, rate-limited, weird playlist) it falls back to the embed scraper
  (today's ≤100). Albums skip the full path (the embed already returns them
  whole). Same return shape either way.
- The two link-path call sites in `web_server.py` now go through
  `fetch_spotify_public`.

**Scope:** this lives entirely in the no-auth `spotify_public_*` path. The
normal credentialed playlist sync is **not touched** — it already pages and
gets full playlists.

## the library is optional + MIT stays clean

SpotipyFree is **GPL-3.0**, so it is **not vendored** into SoulSync's code — it's
declared as a normal **pip dependency** in `requirements.txt` (aggregation,
exactly like spotDL, which is also MIT). The code **soft-imports** it and falls
back to the embed scraper if it's missing, so it's never a hard runtime
requirement and SoulSync's own license stays MIT.

Heads-up: it pulls `spotapi`, `pymongo`, `tls-client`, etc. (~40 MB), and
SpotipyFree needs `websockets` (a transitive dep pip doesn't grab on its own —
added to requirements). After updating, `pip install -r requirements.txt` and
restart.

## tested

- Verified live: pulled all **236/236** tracks of a real playlist via SpotipyFree.
- 9 unit tests: track normalisation, pagination past 100, library-missing →
  raises (→ fallback), and the embed-fallback orchestration (full wins /
  failure → embed / album → embed).
- `ruff check .` clean.

## honest caveats

- The **wiring is bulletproof**: you get the full playlist, or a clean fallback
  to the old ~100 — never a crash, never a partial/truncated-looking list, and
  the authed path is untouched.
- The **data source is best-effort**: it rides Spotify's private API via
  SpotipyFree, so when Spotify shifts their internals the library breaks until
  its maintainer patches it — during that window the link tab quietly returns
  ~100 again, then full once it updates. That's true of any no-login approach.
- Verified one real playlist + the logic; exotic cases (10k-track, collaborative,
  region-locked, lots of local-file/podcast items) degrade to the fallback on
  any error, so worst case is ~100.

### DIFF
--- a/core/spotify_public_api.py
+++ b/core/spotify_public_api.py
@@ -32,7 +32,8 @@ from typing import Any, Callable, Dict, List, Optional
 
 import requests
 
-logger = logging.getLogger(__name__)
+# 'soulsync.*' so these lines land in app.log (the bare module name isn't captured).
+logger = logging.getLogger('soulsync.spotify_public')
 
 _BROWSER_HEADERS = {
     'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 '
@@ -96,6 +97,8 @@ def _paginate_api_tracks(http_get: Callable, spotify_id: str, token: str) -> Lis
             },
             timeout=_TIMEOUT,
         )
+        if resp.status_code != 200:
+            logger.warning("public API tracks offset=%s -> HTTP %s", offset, resp.status_code)
         resp.raise_for_status()
         items = (resp.json() or {}).get('items') or []
         if not items:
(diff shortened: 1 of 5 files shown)


radoslav-orlov:
Now that Spotify requires paid subscriptions for access to its API and by that extent also to be used as enrichment service and metadata source in SoulSync, could SpotipyFree be used instead?

Issue #704, opened 2026-05-26
[Enhancement] AcoustID Retag Ineffective


Tacobell444:
Current Process:

When AcoustID finds an incorrect song it retags the file with the some of the correct information. However, it leaves the file in place and does not rename the file. This creates an issue where the file is now listed with the correct title in Soulsync but it is still on the incorrect Artist/Album in the library in place of the previous track. 

Proposed fix:

When retagging the track, Soulsync should also rename the file and move it to the correct artist/album if it exists or rename it and move it to staging to be imported correctly.

Nezreka:
This is now resolved

PR #781, opened 2026-06-03
docs(readme): document AcoustID API key + retag gotcha

Kartalops:
## Summary
- Expand the AcoustID section of the README with two short notes:
  1. A pointer to <https://acoustid.org/new-application> for the free API key and where to paste it in Settings.
  2. A cross-reference to issue #704 explaining the most common reason the retag action appears to do nothing: the file already carries a cached `MUSICBRAINZ_TRACKID` tag, which the retag path uses as a short-circuit. Removing the cached tag is the documented workaround until the retag path is fixed.

## Why
Issue #704 currently has no body and a thumbs-up from a single user. The retag step is a known UX dead-end for anyone who hits it: the UI takes the "retag" action, nothing visibly changes, and there is no documentation that explains why. Adding a short note to the README gives affected users a definitive answer without a code change.

## How tested
- N/A (docs-only).
- Verified the rendered Markdown in a local preview: no broken links, no formatting issues.

## Checklist
- [x] No new dependencies
- [x] No env-var or behaviour changes
- [x] Does not modify any code path

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,21 @@ SoulSync bridges streaming services to your music library with automated discove
 - Catches wrong versions (live, remix, cover) even from streaming API sources
 - Fail-open design: verification errors never block downloads
 
+#### AcoustID API key
+
+AcoustID verification is opt-in. To enable it, request a free API key
+at <https://acoustid.org/new-application> and paste it into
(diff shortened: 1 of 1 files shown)


Nezreka:
Sounds great! Make sure to set to merge to dev though. We merge dev > main.

PR #786, opened 2026-06-04
Fix wishlist modal: propagate playlist provenance for folder layout

kekkokk:
## Summary

When users add tracks to the wishlist from a **playlist** download modal (“Add to Wishlist”), rows were saved with `source_type: album` and no `playlist_id` / `playlist_name`. On wishlist requeue, `master.py` could not resolve the mirrored playlist or apply **Organize by Playlist** layout, so files defaulted to artist/album paths instead of the playlist folder.

This change:

- Detects playlist vs album download modals (same prefix rules as download/M3U flows).
- Sets `source_type` to `playlist` for playlist modals.
- Stores `playlist_id` (normalized ref for resolve), `playlist_name`, `playlist_source`, and `organize_by_playlist` when the modal toggle is on.
- Keeps album modals unchanged (`source_type: album` + album/artist context).
- Fixes a small bug where missing artists used an undefined `artist` variable.

## Notes

- Existing wishlist rows added before this fix still lack playlist provenance; re-add from the modal or clear and re-import if needed.
- Playlist-folder layout on wishlist retry still requires `organize_by_playlist` enabled on the mirrored playlist in the DB (the modal toggle persists this when starting a download; manual wishlist add now records the intent in `source_info` as well).

## Test plan

- [ ] Sync → open playlist → Download Missing → **Add to Wishlist** → verify `source_info` has `playlist_id` and `playlist_name`
- [ ] With **Organize by Playlist** checked, add to wishlist → process wishlist → files land under `Transfer/<Playlist>/...`
- [ ] Artist album modal → Add to Wishlist → still `source_type: album`, no `playlist_id`
- [ ] Deezer/YouTube playlist modals → `playlist_id` uses normalized ref (`deezer_arl_*` stripped where applicable)

### DIFF
--- a/tests/wishlist/test_routes.py
+++ b/tests/wishlist/test_routes.py
@@ -435,6 +435,41 @@ def test_get_wishlist_cycle_returns_stored_value():
     assert payload == {"cycle": "singles"}
 
 
+def test_add_album_track_to_wishlist_preserves_playlist_modal_context():
+    runtime, service, _db, _logger, _activity_calls = _build_runtime()
+    track = {
+        "id": "track-1",
+        "name": "Song One",
+        "artists": [{"name": "Artist One"}],
+        "duration_ms": 1234,
+    }
+    artist = {"id": "artist-1", "name": "Artist One"}
+    album = {"id": "album-1", "name": "Album One"}
+
+    payload, status = add_album_track_to_wishlist(
+        runtime,
+        track=track,
+        artist=artist,
+        album=album,
+        source_type="playlist",
+        source_context={
+            "playlist_id": "37i9dQZF1DX0XUsuxWHRQd",
+            "playlist_name": "Daily Mix",
+            "organize_by_playlist": True,
+            "added_from": "download_modal",
+        },
+    )
+
+    assert status == 200
+    assert payload["success"] is True
+    add_call = service.add_calls[0]
+    assert add_call["source_type"] == "playlist"
+    assert add_call["source_context"]["playlist_id"] == "37i9dQZF1DX0XUsuxWHRQd"
+    assert add_call["source_context"]["playlist_name"] == "Daily Mix"
+    assert add_call["source_context"]["organize_by_playlist"] is True
+    assert add_call["source_context"]["added_from"] == "download_modal"
+
+
 def test_add_album_track_to_wishlist_builds_spotify_payload_and_merges_context():
     runtime, service, _db, _logger, _activity_calls = _build_runtime()
     track = {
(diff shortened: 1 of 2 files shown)


Nezreka:
hey, thanks for this — the playlist provenance on wishlist adds is exactly the missing piece from the organize-by-playlist work, and that `artist.name` crash fix is a real one (that var doesn't exist in scope, would've thrown).

one thing that'll block the feature for non-spotify playlists though:

you're storing the source as `playlist_source`, but the requeue resolver in `master.py` reads `source`:

```python
# core/downloads/master.py:1047
wl_pl_source = wl_source.get('source') or 'spotify'
```

so `default_source` always falls back to `'spotify'`. for a deezer playlist that means: `normalizePlaylistOrganizeRef` strips `deezer_arl_` down to the bare numeric id, the source lookup misses (wrong source), and it falls through to the `ref.isdigit()` PK branch in `resolve_mirrored_playlist` — which either finds nothing or resolves a totally unrelated playlist by primary key. that's the exact thing that resolver's docstring says it was written to avoid. spotify only works here by accident.

easy fix — just emit `source` instead of (or alongside) `playlist_source` in `buildModalWishlistSourceContext`. that's the key the rest of the pipeline uses everywhere, including where master.py rewrites source_info a few lines down.

couple smaller things, not blockers:

- the classifier is a denylist (anything not in the album/single/redownload prefix lists becomes a playlist). the `resync_*` album-resync modal id isn't in there, so an album resync added to wishlist gets tagged as a playlist with a junk id. harmless in practice (requeue misses and falls back to the album path) but it stores a misleading row + drops the album context keys. would lean toward an allowlist of the real playlist prefixes so it fails safe to album.
- fyi `ui_playlist_ref`, `added_from`, `timestamp` and the `organize_by_playlist` flag inside source_context aren't read on requeue — organize gets re-derived from the live mirrored_playlists row. fine as provenance, just heads up they don't drive anything.
- the new test only covers the backend pass-through, which already worked. the actual new bits (`resolveModalWishlistSourceType` / `buildModalWishlistSourceContext`) aren't tested — a quick seam test on the classifier (`deezer_arl_x` -> playlist, `artist_album_x` -> album, `resync_x` -> album) would've caught the source-key thing.

fix the `source` key and i'm happy to merge. nice work otherwise 🙏

ramonskie:
any idea when this gets merged?
as currently playlists are not mapped to a folder

ramonskie:
you can cherry pick https://github.com/ramonskie/SoulSync/commit/84660a592486bfdca2ce4d67e445baa978674ef1
to fix this pr

tested this localy

ramonskie:
can be closed since the refactor in 2.7.2

Issue #797, opened 2026-06-05
AcoustID will always fail with non-english artists


Tacobell444:
AcoustID always fails and quarantines downloads from originally non-english artists. (eg. Joe Hisaishi) because AcoustID reports metadata in the original language and soulsync is comparing with english metadata. 

I'm not sure of a graceful fix for this. Maybe there's a way to match artist/song aliases in other languages? Or adding a toggle to enable/disable acoustID to the album request flow? Like where the "Force Download All" toggle is?

Nezreka:
Good ideas, I have this live on dev now. Will push main soon

Tacobell444:
Sorry, this is still falsely quarantining correct tracks. 

<img width="556" height="558" alt="Image" src="https://github.com/user-attachments/assets/9eef0c46-b217-4c52-a5df-d19f662bd203" />

<img width="1486" height="853" alt="Image" src="https://github.com/user-attachments/assets/b1e0bcbd-4a8a-44de-b4e8-f3aa7b7ee738" />

PR #801, opened 2026-06-05
Downloads: complete retry overhaul,  exhaustive multi-source retry, MusicBrainz kanji fix, version-mismatch last-resort fallback

nick2000713:
## What problem does this solve?

  SoulSync's download retry system had several independent bugs that caused tracks to fail permanently even when correct files were available. This PR fixes the **entire retry pipeline end-to-end**, adds a **kanji/romaji artist 
  recognition fix**, and introduces an **opt-in last-resort fallback** for tracks that simply don't exist in the wanted version on any source.

  All three subsystems were confirmed broken in production with an Attack on Titan OST test corpus (Sawano Hiroyuki / KOHTA YAMAMOTO), and all three are confirmed fixed by production logs after this PR.

  > **Note on defaults:** Every new retry behaviour introduced here can be tuned or disabled in Settings → Downloads → Retry Logic. With exhaustive retry off and the version-mismatch fallback off, the system behaves identically to the
  old default — all new features are strictly opt-in or configurable.

  ---

  ## 1 — Retry pipeline overhaul

  ### The old behaviour (broken)

  When a downloaded file failed AcoustID or integrity verification, SoulSync quarantined it and re-queued the task. But the retry logic had three independent bugs:

  **Bug A — source abandoned after one query.**
  The old code tracked retries at the *source* level. Soulseek has dozens of peers and six different search queries per track. After the first peer returned the wrong file, the entire Soulseek source was marked exhausted and the task
  switched straight to HiFi — leaving 100+ potentially-correct Soulseek peers untried.

  **Bug B — exhausted source = hard failure.**
  When a source's retry budget ran out, the task failed outright. It should switch to the next configured fallback source (e.g. Soulseek → HiFi) instead.

  **Bug C — no per-source retry budget.**
  By default, only one global retry counter existed. There was no per-source budget that scales with how many search variants exist for a track.

  ### The new behaviour

  Retries are now tracked at the **query level** within each source:

  Track: "Call Your Name (Gv)" by Hiroyuki Sawano
  Retry budget (exhaustive mode): 6 queries × 5 retries/query = 30 attempts per source

  Attempt 1/30  → soulseek peer A returns Instrumental → quarantine → retry
  Attempt 2/30  → soulseek peer B returns Instrumental → quarantine → retry
  ...
  Attempt 12/30 → soulseek peer L returns Instrumental → quarantine → retry
  → Switch to HiFi (soulseek budget exhausted)
  HiFi: no valid candidates found
  → Last-resort fallback kicks in (see section 3)

  Already-searched queries are skipped on re-entry; not-yet-searched queries against the same source are still attempted. The cached-first fast path (walk already-found candidates before re-searching) is preserved.

  ### New settings (Settings → Downloads → Retry Logic)

  | Setting | Default | Description |
  |---|---|---|
  | **Retry next-best candidate** | on | Re-queue failed downloads; try the runner-up peer instead of giving up immediately |
  | **Exhaustive retry** | off | Enable per-source budgets (query × retries/query); off = one global 5-attempt cap |
  | **Retries per query** | 5 | How many peers to try per search query before moving on. Budget = queries × this value |

  ---

  ## 2 — MusicBrainz kanji/romaji artist fix

  ### The bug
  
  AcoustID verification compares the track's embedded artist tag against the artist expected by the playlist. For Japanese artists whose name exists in both kanji (澤野弘之) and romaji (Hiroyuki Sawano / Sawano Hiroyuki), the library
  uses MusicBrainz aliases to bridge the scripts.

  The alias lookup (`lookup_artist_aliases`) has a "MB-score escape hatch": if a MusicBrainz result has a near-perfect raw MB score (≥ 95, unambiguous), use it even if the string-similarity to the query is low — because the name is in a
  different script.

  The bug: the escape hatch checked the MB score of the **combined-score leader** (highest string-similarity × MB score) instead of the **MB-score leader** (highest raw MB score). For cross-script lookups these are different entities:

  Query: "Sawano Hiroyuki"
                            String-sim   MB score   Combined
  SawanoHiroyuki[nZk]  →     0.82         83         68.1  ← combined leader (WRONG pick)
  澤野弘之              →     0.30        100         30.0  ← MB-score leader (CORRECT pick)

  The combined leader's MB score was 83 (< 95 threshold) → escape hatch rejected → aliases returned `[]` → every file by Sawano failed AcoustID with `artist=0%`.

  ### The fix

  The escape hatch now independently finds the entity with the highest **raw MB score** and tests that entity's score against the threshold. The combined-score leader is still used for the normal path; the MB-score leader is only used
  when the escape hatch fires.

  Result: 澤野弘之 selected → aliases include "Hiroyuki Sawano" → artist=100% ✓

  Confirmed in production: Barricades, YouSeeBIGGIRL/T:T, APETITAN, Call of Silence all pass with `artist=100%` after deploy + cache purge.

  > **Note:** a stale `artist_aliases` cache entry (30-day TTL) will block the fix until purged. Safe one-time purge:
  > ```sql
  > DELETE FROM musicbrainz_cache WHERE entity_type='artist_aliases';
  > ```

  ---

  ## 3 — Version-mismatch last-resort fallback

  ### The problem

  Some tracks simply don't exist on any configured source in the wanted version. Example: "Call Your Name (Gv)" — every copy on Soulseek and HiFi is the *Instrumental* version. The retry system correctly rejects each one (version
  mismatch), exhausts all 30 attempts across two sources, and leaves the track permanently missing.

  There is a second, subtler case: AcoustID's own database can be wrong. A file may be perfectly correct — right song, right artist, right cut — but AcoustID has it misidentified or tagged with the wrong version label. In that situation
  the verification system will reject a correct file indefinitely, and no amount of retrying will ever produce a "passing" candidate because every copy of the track fingerprints identically. The fallback handles this case too: after N
  consistent rejections all citing the same mismatch, it accepts the best candidate rather than leaving the track missing forever.
  
  ### The solution (opt-in)

  Once retries are fully exhausted, if every quarantined candidate for this track failed with **the same version mismatch** (e.g. all instrumental) and there are at least N of them, accept the best one rather than failing. "Best" =
  oldest/first-tried (highest-confidence match from the original search ranking).

  **Hard safety rules — none of these are relaxed:**
  - Only `Version mismatch` quarantine entries qualify. Audio/artist mismatches (genuinely wrong song) and integrity/duration failures (truncated/corrupt file) never participate.
  - All qualifying entries must share the **same matched version**. A mix (instrumental + live) is ambiguous → no acceptance.
  - The accepted file is re-imported with **only the AcoustID gate bypassed**. Integrity, duration, and bit-depth checks still run — a truncated or corrupt file is still caught.

  **Two trigger paths** — the fallback fires when:
  1. AcoustID retries are exhausted after repeated post-processing failures (`pipeline.py`)
  2. The retry *search* itself finds zero valid candidates — the source returns nothing at all (`task_worker.py`)

  ### New settings (Settings → Downloads → Retry Logic)
  
  | Setting | Default | Description |
  |---|---|---|
  | **Accept best version mismatch as last resort** | off | Enable the fallback — must be explicitly turned on |
  | **Min. same-version count** | 2 | How many quarantined entries with the same mismatch are required before accepting |

  ### Confirmed working in production
  
  [Version-Mismatch Fallback] retries exhausted for 'Hiroyuki Sawano - Call Your Name (Gv)';
  accepting best quarantined candidate (instrumental, entry 20260605_235536_09 - Call Your Name (Gv)) as last resort

  ---

  ## UI changes

  - **New collapsible tile** "Retry Logic" under Settings → Downloads — consolidates all retry settings (previously scattered under Post-Processing)
  - **New collapsible tile** "Quality Profile" under Settings → Downloads — moved from its previous location; gated by Soulseek being an active source
  - **Sidebar hover restored** in reduce-visual-effects mode — the previous performance PR accidentally removed hover highlights from the sidebar nav buttons

  ---
  
  ## Tests

  | File | What's covered |
  |---|---|
  | `tests/downloads/test_downloads_task_worker.py` | Lazy query-level retry: unsearched query still hits same source; searched query skipped; budget-exhausted source excluded; queries recorded; stale state cleared on non-quarantine run
  |
  | `tests/imports/test_version_mismatch_fallback.py` | `select_version_mismatch_fallback`: min-count gate, mixed-version ambiguity rejection, single-version acceptance, oldest-first selection; `try_accept_version_mismatch_fallback`:
  setting disabled, approve failure, context mutation, reprocess dispatch |
  | `tests/imports/test_import_pipeline.py` | Full pipeline integration: fallback wired at AcoustID give-up point |
  | `tests/matching/test_artist_alias_lookup_586.py` | MB trust-gate uses MB-score leader: cross-script artist correctly selected when combined-score leader has lower raw MB score |

  ---

  ## Files changed

  | Area | Files |
  |---|---|
  | Core retry logic | `core/downloads/monitor.py`, `core/downloads/task_worker.py` |
  | Import pipeline + fallback | `core/imports/pipeline.py`, `core/imports/version_mismatch_fallback.py` |
  | MusicBrainz fix | `core/musicbrainz_service.py` |
  | Settings | `config/settings.py` |
  | WebUI | `webui/index.html`, `webui/static/settings.js`, `webui/static/style.css` |
  | Wiring | `web_server.py` |
  | Tests | `tests/downloads/`, `tests/imports/`, `tests/matching/` |

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -494,6 +494,19 @@ class ConfigManager:
                 "album_bundle_poll_interval_seconds": 2.0,
                 "album_bundle_timeout_seconds": 6 * 60 * 60,    # 6 hours
             },
+            "post_processing": {
+                # When a download is quarantined (AcoustID mismatch, integrity /
+                # duration failure), retry the next-best candidate instead of
+                # failing outright. Default on.
+                "retry_next_candidate_on_mismatch": True,
+                # Opt-in exhaustive retry: budget retries PER SOURCE so every
+                # source (Soulseek, then HiFi/Tidal/…) gets its own attempts
+                # before the track gives up. Default off (single global cap).
+                "retry_exhaustive": False,
+                # Retries per search query per source in exhaustive mode. The
+                # per-source budget is query_count × this value.
+                "retries_per_query": 5,
+            },
             "tidal_download": {
                 "quality": "lossless",  # Options: "low", "high", "lossless", "hires"
                 "session": {
--- a/config/settings.py
+++ b/config/settings.py
@@ -497,8 +497,8 @@ class ConfigManager:
             "post_processing": {
                 # When a download is quarantined (AcoustID mismatch, integrity /
                 # duration failure), retry the next-best candidate instead of
-                # failing outright. Default on.
-                "retry_next_candidate_on_mismatch": True,
+                # failing outright. Default off — opt-in.
+                "retry_next_candidate_on_mismatch": False,
                 # Opt-in exhaustive retry: budget retries PER SOURCE so every
                 # source (Soulseek, then HiFi/Tidal/…) gets its own attempts
                 # before the track gives up. Default off (single global cap).
--- a/core/downloads/monitor.py
+++ b/core/downloads/monitor.py
@@ -237,6 +237,12 @@ def requeue_quarantined_task_for_retry(task_id, batch_id, trigger):
         task['used_sources'] = used_sources
 
         task['quarantine_retry_count'] = total_count + 1
+        # Flag the re-run as a quarantine retry so the worker walks the
+        # already-found candidates (cached-first) before re-searching — the
+        # connection was fine, the content was just wrong. Dead-connection /
+        # stuck retries (handled elsewhere in the monitor) deliberately do NOT
+        # set this, so they re-search fresh.
+        task['_quarantine_retry'] = True
         # Drop the stale download identity + the prior attempt's quarantine link.
         task.pop('download_id', None)
         task.pop('username', None)
--- a/core/downloads/monitor.py
+++ b/core/downloads/monitor.py
@@ -37,9 +37,98 @@ missing_download_executor = None
 download_orchestrator = None
 _RELEASE_SOURCE_NAMES = frozenset(('torrent', 'usenet'))
 
+# Hard ceiling on automatic next-candidate retries after a download was
+# quarantined (AcoustID mismatch / integrity / duration). The natural
+# terminator is used_sources exhaustion — once every candidate the worker can
+# find has been tried, attempt_download_with_candidates returns False and the
+# worker reports a clean failure. This cap is a safety net against a pathological
+# quarantine→retry→quarantine loop (e.g. a source that keeps returning fresh
+# wrong files).
+MAX_QUARANTINE_RETRIES = 5
+
 
 def _download_id_key(download_id):
     return f"download_id::{download_id}" if download_id else None
+
+
+def requeue_quarantined_task_for_retry(task_id, batch_id, trigger):
+    """Re-queue a task whose download was just quarantined so the worker tries
+    the NEXT best candidate instead of failing outright.
+
+    Called from the post-processing verification wrapper when AcoustID
+    verification or the integrity/duration check quarantines a file. It mirrors
+    the monitor's transfer-error retry path: mark the bad source as used, clear
+    the stale download identity, reset the task to ``searching`` and resubmit
+    the download worker. Because ``used_sources`` is preserved across the
+    re-run, the worker skips the quarantined source and picks the next-best
+    candidate (see ``attempt_download_with_candidates``).
+
+    Returns True if a retry was queued — the caller must then NOT mark the task
+    failed or notify batch completion, since the task is going around again.
+    Returns False when no retry is possible (retry engine unwired, manual pick,
+    cancelled, or retry budget exhausted); the caller falls through to its
+    existing failure handling.
+    """
+    # Opt-out escape hatch — default on. Lets users restore the old
+    # quarantine-and-fail behaviour without a code change.
+    if not config_manager.get('post_processing.retry_next_candidate_on_mismatch', True):
+        return False
+
+    # Retry engine not wired (e.g. manual-import path that never started a
+    # download worker). Nothing to re-run.
+    if missing_download_executor is None or _download_track_worker is None:
+        return False
+
+    with tasks_lock:
+        task = download_tasks.get(task_id)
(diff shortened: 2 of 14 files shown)


nick2000713:
Hi! I noticed that #797 (just merged) and this PR both tackle the same
cross-script artist verification problem from different angles. I want
to explain why both fixes are complementary and why this PR's
musicbrainz_service.py change is still needed even with #797 in place.

**#797's approach (acoustid_verification.py)**
When the expected artist and the AcoustID-matched artist span writing
systems AND the artist is confirmed via the MusicBrainz alias bridge
→ skip quarantine.

**This PR's approach (musicbrainz_service.py, commit 0595db0d)**
Fixes the alias bridge itself. The trust gate in `lookup_artist_aliases`
was evaluating the MB score of the *combined-score leader* instead of
the *MB-score leader*. For cross-script artists these are different
entities:

Query: "Sawano Hiroyuki"
                         String-sim   MB score   Combined
SawanoHiroyuki[nZk]  →    0.82         83         68.1  ← combined leader
澤野弘之              →    0.30        100         30.0  ← MB-score leader

Without this fix, `lookup_artist_aliases("Sawano Hiroyuki")` returns
`[]` → `artist_sim = 0%` → artist is NOT confirmed → **#797's
cross-script skip never triggers** either.

With both fixes: the alias lookup returns the correct aliases
→ `artist_sim = 100%` → artist confirmed → #797's skip also works
as a second layer of protection.

Confirmed in production: Sawano Hiroyuki (澤野弘之) tracks —
Barricades, YouSeeBIGGIRL/T:T, APETITAN, Call of Silence — all pass
with `artist=100%` after deploying this fix.

This PR has been rebased onto the current dev tip (includes #797).

---

PR #845, opened 2026-06-10
feat: import folder-artist override opt-in + verification pipeline review queue

nick2000713:
## What this does

  Two things that came out of running SoulSync on a large multi-artist library.

  ### 1. Import: folder-artist override is now opt-in

  `auto_import_worker` was unconditionally overriding the matched album-artist with the parent Staging folder name.
  On a multi-artist library this produces a mess — every track that comes through a folder named "soulsync" or
  "downloads" gets filed under that as the artist. The matched artist from the pipeline gets thrown away.
  
  The override is now gated behind `import.folder_artist_override` (default `on`). When it's off, the pipeline's
  matched artist is kept. Turning it on restores the original behaviour for setups that rely on it.

  > I'm not sure this setting lives in the right place — happy to move it or rename the key if you have a
  preference.

  ### 2. Verification review queue on the Downloads page

  A new **"⚠ Unverified / Quarantine"** filter pill on the Downloads page. Two sub-views:

  **Unverified** — tracks that were imported but didn't pass AcoustID. Per-row actions: play the local file, inline
  stream-search comparison (same pipeline as /search, no new tab), open the Audit Trail, delete. Bulk delete for the
  whole list.
  
  **Quarantine** — the existing quarantine entries, now with cover art, an inline inspect panel (sticky across the
  2-second polling re-render), and an Audit Trail modal reachable via the 🔍 button. Approving a quarantined file
  sets `human_verified` status and re-triggers import.
  
  If `acoustid.enabled` is off, the Unverified sub-view is hidden and the pill label collapses to "🛡 Quarantine" —
  resolved eagerly on page load, not after the first click.

  **Scanner (`acoustid_scanner.py`):** PASS writes `verified` to tag + DB; SKIP writes `unverified` and inserts a 
  history row if none exists, so the file surfaces in the review queue after a scan. Removed `skip_force_imported` —
  only `human_verified` short-circuits a rescan now.
  
  **New tag:** `SOULSYNC_VERIFICATION` (Vorbis / ID3 TXXX / MP4 freeform) — survives DB resets.

  > Same question here: is the Downloads page the right home for this, or would a dedicated page be better? Any 
  preference on the tag name?

  ## Live verification
  
  Tested end-to-end on a real library: folder-artist fix stopped ~200 mis-filed tracks, quarantine approve correctly
  sets `human_verified` and imports, the scanner populates the unverified queue after a scan, and the pill 
  correctly shows "🛡 Quarantine" on fresh page load when verification is disabled.

  ## Would appreciate your eyes on

  - The folder-artist fix changes import control flow — worth a real run with `use_folder_as_artist` both on and off
  - Quarantine approve → `human_verified` → re-import, to make sure it fits cleanly with your quarantine handling
  - The review queue placement — if Downloads isn't the right page I'd rather know before building more on top of it
  - Test coverage: added tests for the scanner status persistence and the folder-artist gate — let me know if there 
  are existing modules I should hook into instead


Nezreka:
This looks fantastic! I'll test it thoroughly soon but it looks like there is a merge conflict XD

nick2000713:
Just wanted to give you a full update before I step back for my exams (this week and next week).

**Merge conflict** is resolved — I pushed a fix for the `style.css` conflict earlier today.

**Regarding issue #851 (forward slashes in titles):** I actually just tried to download "You See Big Girl / T:T" by Sawano Hiroyuki and ran into this exact bug. The search DID find 9 candidates for the query `You See Big Girl / T:T`, but all of them were rejected by the candidate matcher.

To clarify where the two pieces are:
- **Search query generation** works fine overall — it generates 5 queries, some normalized (e.g. `you see big girl tt`) and some raw (e.g. `You See Big Girl / T:T`). Soulseek handles both and returns results either way.
- **Candidate matching** is where it breaks. The files on Soulseek are named `You See Big Girl _ T_T.flac` (underscores), and when the similarity check in `core/matching/audio_verification.py` compares `You See Big Girl / T:T` against that filename, the `/` and `:` aren't normalized before comparison, causing the score to drop below the threshold and all 9 candidates to be rejected.

This PR introduces `core/matching/audio_verification.py` as the single source of truth for title/artist normalization — both import-time verification and the library scan now share one definition. That's exactly why this is the right place to fix it: the fix should go in the `normalize()` function there, replacing `/` and `:` with spaces before the similarity comparison, similar to how `<>`, `()` and `[]` are already handled. No need to fix it in multiple places anymore.

I'd leave this for you to handle as part of issue #851 — I don't want to pile on more changes while I'm in exam mode and risk more conflicts.

PR #848, opened 2026-06-11
Discogs: fix master/release ID collision fetching the wrong album

### LINKED ISSUE #847: Discogs metadata returns the wrong album (master/release ID collision)
First off — thank you for SoulSync, it's already incredibly useful and I really appreciate all the work in it. 🙏 I'm a DJ and a lot of what I dig for is obscure
  material that realistically only Discogs has proper metadata for, so rock-solid Discogs integration would make SoulSync **super** helpful for me. While leaning on
  the Discogs source I hit a consistent bug:

  ## Summary

  When I search using Discogs and SoulSync then fetches further details for a result, the data comes back for a *different* album. The ID it uses is always valid — it
  just points at the wrong object.

  ## Root cause

  Discogs has two album object types that share one numeric ID space:

  - releases — `/releases/{id}`
  - masters — `/masters/{id}`

  Release `N` and master `N` are different albums. SoulSync's album search uses `type=release`, so it gets **release** IDs — but the follow-up fetch (album details /
  tracklist / enrichment) tries `/masters/{id}` first and only falls back to `/releases/{id}`. Whenever a release ID also happens to be a valid master ID (common,
  since masters number in the millions), the master lookup succeeds and returns a valid-but-wrong album, so the release fallback never runs.

  ## Steps to reproduce

  1. Configure a Discogs token and use Discogs as the metadata source.
  2. Search for an album and select a result.
  3. The fetched album details / tracklist belong to a different release than the one selected.

  **Expected:** details/tracklist match the selected album.
  **Actual:** a different album is returned (valid ID, wrong object). Because the IDs collide silently it affects many ordinary searches, not just edge cases.

  ## Impact

  Affects the Discogs metadata/search source and the background Discogs enrichment worker (which can store the wrong release's genres/labels/track count against a
  library item).

  I have a fix and will open a PR against `dev` referencing this issue. Thanks again!

RollingBase:
Fixes #847 

  ## Problem

  Discogs masters (`/masters/{id}`) and releases (`/releases/{id}`) share a single numeric ID space, so release `N` and master `N` are different albums. Album search
  returns **release** IDs, but the follow-up detail/tracklist/enrichment fetch tried `/masters/{id}` first with a `/releases/{id}` fallback. When a release ID is also
  a valid master ID (common), the master lookup succeeds and returns a valid-but-wrong album — the fallback never runs.

  ## Fix

  - **Type-tag album IDs at the parse boundary** — IDs are prefixed `r<id>` (release) or `m<id>` (master) when first read from Discogs, so the object type travels
  with the ID.
  - **Route each fetch to the correct endpoint** based on that tag — no more master-first guessing.
  - **Legacy bare IDs self-heal** — untagged numeric IDs are resolved release-first (then master fallback) and re-stored tagged. No DB migration needed.
  - The background enrichment worker persists tagged IDs too.

  ## Tests

  Added `tests/test_discogs_id_typing.py` (21 tests) covering endpoint routing, kind detection, ID tagging, `Album.from_discogs_release` tagging, and a regression
  lock that a release ID never hits `/masters`.

  ## Verification

  Ran the CI gates locally (Python 3.11, `requirements-dev.txt`):

  - `ruff check .` — clean
  - `compileall` — ok
  - `pytest` — 5656 passed


RollingBase:
I see that my fix isn't yet fully working locally, I'll keep iterating untill I have something working! :)

RollingBase:
The issue was with images and data from discogs not populating. This issue seems to be preexisting, so I created a separate issue for it #849 . This branch should fix the discogs loading as is! :)

PR #853, opened 2026-06-11
Cache artist album lists across metadata sources

ramonskie:
## Summary
- add shared artist album-list cache helper for metadata providers
- use it from Spotify, Deezer, iTunes, and Discogs while preserving Spotify truncated-fetch protection
- add regression coverage for non-Spotify cache reuse and existing Spotify cache-poisoning behavior

## Why this cache is needed
The artist-detail page fetches a full source-priority discography before it can hide `Loading artist discography...`. Before this change, Spotify was the only provider with an artist album-list cache. Deezer, iTunes, and Discogs cached individual album entities, but not the artist → album-list response, so reopening the same library artist could refetch the full remote discography every time. Because the default source priority tries non-Spotify providers before Spotify, users could still see slow artist loads even when Spotify caching existed.

This PR moves album-list caching into a shared helper and applies it consistently across the metadata clients. Spotify keeps its existing complete-discography cache behavior and the truncated-fetch guard so partial watchlist probes cannot poison the artist-detail cache.

## Tests
- python -m pytest tests/test_artist_album_list_cache_nonspotify.py tests/test_artist_albums_cache_poisoning.py
- python -m ruff check core/metadata/artist_album_cache.py core/spotify_client.py core/deezer_client.py core/itunes_client.py core/discogs_client.py tests/test_artist_album_list_cache_nonspotify.py tests/test_artist_albums_cache_poisoning.py

### DIFF
--- a/core/deezer_client.py
+++ b/core/deezer_client.py
@@ -6,6 +6,7 @@ from typing import Dict, List, Optional, Any
 from functools import wraps
 from dataclasses import dataclass
 from utils.logging_config import get_logger
+from core.metadata.artist_album_cache import get_cached_artist_album_items, store_artist_album_items
 from core.metadata.cache import get_metadata_cache
 
 logger = get_logger("deezer_client")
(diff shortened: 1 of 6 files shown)


ramonskie:
damn a shame that i missed https://github.com/Nezreka/SoulSync/commit/826ac0b3664e556365ddf53a02537d602830d6f3

Issue #652, opened 2026-05-20
Quarantined Files Being Duplicated


S-Bryce:
As a part of the quarantine process, files are stored for later review by me. However, it seems that future automatic download attempts are not considering whether they have already downloaded the same file before, resulting in me waking up to potentially hundreds of duplicate quarantined files:

<img width="677" height="1137" alt="Image" src="https://github.com/user-attachments/assets/d2f583a2-ee4e-446f-b910-c7cf7a902e18" />

As shown in the screenshot above, this is also not a result of the download system attempting alternative downloads - it is downloading from the same URL and create an exact copy.

While the quarantine system itself appears to be working (notwithstanding the inaccuracy of the acoustid matching), but deduplication does not appear to be getting either checked or run. This, combined with an inability to do bulk or multi-select options on quarantines, results in needing to manually clear the `ss_quarantine` folder regularly.

Nezreka:
I have this fixed on dev. will push main soon

PR #896, opened 2026-06-20
Global quality system: real-audio verification, best-quality search & quality profiles (please try...not ready to merge)

nick2000713:
## ⚠ Please don't merge this yet
  I'd really like you to pull it, run it for a while, and tell me what you think. I'm **not fully comfortable with it yet** and it needs more real-world testing
  before it should actually go in. Consider this a "try it and give me your opinion" PR, not a merge request.

  ---

  This is a fairly large body of work that turns quality from a loose, source-by-source guess into a single, verified, profile-driven system across **search → 
  download → import → library**.

  ### Global quality model & profile
  - New global `AudioQuality` model — one source of truth for what a file *actually* is (format, bit depth, sample rate, bitrate), populated from real probing
  rather than the source's claim.
  - **v3 ranked-target quality profile** with a drag-to-reorder UI editor: you define an ordered ladder of acceptable targets and the system honours that order.
  - Per-source download tier is now **derived from the global profile** instead of being configured separately, so everything stays consistent.
  - Audio quality strings now show sample rate too (e.g. `FLAC 24bit/96kHz`).
  - "Audiophile" preset is now **truly 24-bit-only** (no silent 16-bit/MP3 fallthrough).
  - **Per-preset memory**: edits you make to a preset are remembered when you switch away and back, plus a **"Reset to defaults"** action per preset.

  ### Real-audio verification (the actual quality check)
  - A real **ffmpeg-based audio guard** runs post-download *and* in the library scanner — same core, so a file is judged the same way everywhere.
  - **30-second preview / silence detection**: mostly-silent or 30s preview/sample files are caught and rejected instead of being imported as the real track.
  - **Truncated-download detection**: files whose real decoded audio is shorter than the container claims are flagged.
  - HiFi sources that return a **30s preview manifest** are now declined so we fall through to a real source.
  - Quality gate runs **before** AcoustID, and a quality/audio-guard quarantine no longer wrongly marks the task as completed.

  ### Best-quality search mode
  - New opt-in `best_quality` search mode: instead of taking the first source that meets a target, it **pools candidates from every source per query** and
  downloads the highest *actual* audio quality, with source order only breaking ties.
  - Default `priority` mode is **byte-for-byte unchanged**; retry budgets are untouched.
  - Clearer fallback UI; removed the confusing per-source "Allow quality fallback" checkboxes.

  ### Library quality check as a repair job
  - The old wishlist-style scanner is reworked into a **findings-based repair job**: it reports problem files and lets you act (delete / redownload / ignore).
  - Runs the same real-audio guard as downloads, scopes correctly to the Music Library folder (not the downloads dir), and resolves relative library paths
  reliably (with diagnostics for files it can't probe).

  ### Import-time quality filtering
  - Master toggle to quality-filter on import, surfaced as a collapsible tile, plus the quality-filter / folder-artist toggles on the Import page.

  ### Downloads & quarantine UX
  - Quarantine review UI shows the **real** measured audio quality.
  - Terminal rows (completed/failed/unverified) stay visible during an active batch instead of being starved out by active tasks.
  - Unverified review actions always load; the Downloads page surfaces the full unverified history; stuck post-processing tasks no longer phantom-complete.

  ---

  ### Status / testing
  - Built and running on my own server (amd64); core flows exercised by hand, automated test suite is green (quality / imports / downloads / repair jobs).
  - Still wants broader real-world testing — that's exactly why I'm asking you to try it first. **Feedback very welcome before we even think about merging.**


nick2000713:
hey, since quality is now a global thing that affects downloads, imports, library scanning and repair jobs i was thinking it might make sense to give it its own tab in settings? right now the quality profile and the acoustid behaviour (require verified, skip handling etc.) are under downloads, and the import quality gate ("only import tracks that meet your profile", replace lower quality) is buried under library → import. would be nice to have it all in one place so it's clear this is a system-wide concept and not just something per source like it used to be. what do you think?

nick2000713:
hey @Nezreka, just wanted to give a quick status update... things are running pretty well on my end, the core flows all work and i've been using it on my own server for a while now. the one thing still open on the ui side is  whether it makes sense to move the quality settings into their own dedicated tab in settings, since it's now a global system that touches downloads, imports and library scanning..happy to do that if you think it's a good idea, just didn't want to do it without a heads up.

also noticed issue #921, that's basically exactly what this PR solves. the quality profile now has a full drag-to-reorder target ladder so users can set any priority order they want (mp3 320 above flac, or whatever makes sense for them).

nick2000713:
@Nezreka, I know the Quality Upgrade Finder still has some issues ...the path resolver currently returns None for all library tracks so no files get probed and 0 findings come out. The auto-delete of the old file after a successful import is also not implemented yet.

That said, I'd like to merge this now... the branch has grown quite large and I'm comfortable with everything else (quarantine UI consolidation, best-quality search mode, boolean settings fix for repair jobs). I'll revisit the Quality Upgrade Finder in a separate branch. Also once you've had a look.. whether the quality settings should get their own dedicated tab in the UI.

Nezreka:
first off — this is seriously impressive work. it is really clean: the new `core/quality/` engine, the ranked-target model, the guard fail-open/opt-in posture, and the frontend (the wishlist-tools stuff getting re-homed to the downloads review queue instead of just deleted) all read well. thank you for putting this together, especially mid exam season.

i did a deep pass over the whole diff and actually ran the code/tests against the branch. there's a lot to like, but there are two hard blockers that have to get fixed before this can merge, plus a few things i want to confirm are intended. laying it all out below.

---

## 🔴 blockers (must fix before merge)

### 1. the app won't boot — db init crashes on every existing install

`database/music_database.py` — the `CREATE INDEX ... ON library_history (verification_status)` runs *before* the `verification_status` column gets added (the `ALTER TABLE` loop is a few lines below it, and the `CREATE TABLE IF NOT EXISTS` doesn't include the column).

i spun up a fresh `MusicDatabase()` and it throws:

```
OperationalError: no such column: verification_status
```

…and init aborts. this hits **fresh installs AND every existing user's db** (any db whose `library_history` table predates this column, which is all of them). i'm pretty sure it works on your machine because your dev db already picked up the column somewhere along the way — but on a clean checkout it dies on startup.

**fix:** move that `CREATE INDEX idx_lh_verification_status` down to *after* the column-migration loop (put it next to `idx_lh_origin`, which is already correctly placed after its `ALTER TABLE`s).

### 2. the #652 quarantine loop is back

the rewritten `filter_results_by_quality_preference` in `core/soulseek_client.py` doesn't call `_drop_quarantined_sources` anymore. that pre-filter is what stops a previously-quarantined `(user, file)` from winning the picker again, re-downloading, re-failing acoustid, and re-quarantining forever (#652).

the existing pinned test catches it — it fails on this branch:

```
tests/downloads/test_soulseek_pinning.py::test_filter_results_by_quality_runs_quarantine_dedup_first
AssertionError: Quarantined source must be filtered before the quality picker
```

**fix:** re-add the `_drop_quarantined_sources(results)` call at the top of the new method (the method's still there at the old spot, it's just not wired in anymore).

---

## 🟠 want to confirm these are intended

### 3. the default search path is now quality-gated for everyone

`core/download_engine/engine.py` `search_with_fallback` always ranks every source's results against the profile, not just in best-quality mode. the issue: an mp3 with `bitrate=None` (slskd leaves bitrate off a *lot* of mp3 results) comes back `satisfied=False`, so soulseek mp3-only results get deprioritized and it'll try other sources first. `fallback_enabled` still completes the download, so nothing breaks — but it changes which source wins and adds latency for default/priority-mode users who didn't opt into anything.

i reproduced `satisfied=False` for `mp3 bitrate=None` against the default migrated profile. if this is the intended new behavior, cool — but it's a real change to the default flow. if not, gate it behind `best_quality` mode so priority-mode stays byte-for-byte the same.

### 4. metadata-less flac over-claims the top tier

`core/quality/model.py` `matches_target` — a flac with no sample_rate / bit_depth metadata (super common on slskd results) matches a strict 24/192 target, while a *real* 16/44 flac doesn't:

```
bare flac        -> matches 24/192 target: True
real 16/44 flac  -> matches 24/192 target: False
```

so under a hi-res / audiophile profile, an unknown-metadata flac can outrank and discard a genuine cd-quality flac. the default `flac (any)` profile isn't affected (no strict constraints), so this is scoped to hi-res targets — but it's backwards. an unknown spec shouldn't satisfy a constraint that requires a specific minimum; it should fail the high tiers and fall to a plain flac bucket.

### 5. per-source hi-res preference looks like it gets silently downgraded

the per-source quality dropdowns are gone and the clients now pull their tier from the global profile via `quality_tier_for_source`. i don't see a migration that seeds hi-res targets from an existing `tidal_download.quality='hires'` / `qobuz.quality='hires_max'`. so a user who had tidal/qobuz on hires would quietly drop to lossless until they manually re-add a hi-res target. can you double check this? if there's no migration, we either need one or at least a heads-up in the release notes.

---

## 🟡 lower priority / nits

- **the branch ships with a red test suite.** `tests/repair_jobs/test_quality_upgrade.py` still calls `meets_preferred_quality` / `classify_track_quality` / `preferred_quality_floor` / `RANK_*`, which this PR deleted from `quality_upgrade.py` — 16 failures. only tests reference them (no runtime break), but they need updating/removing so CI is green. worth a full `pytest` pass before the next push — that's also what would've caught blockers #1 and #2.
- **`.aiff` is probed with the wrong mutagen class** (`mutagen.wave.WAVE` instead of `mutagen.aiff.AIFF`) in `core/imports/file_ops.py`. it fails open so it doesn't reject anything, but it means aiff silently bypasses the quality filter.
- with `fallback_enabled` **off**, the quality guard rejects aac/ogg/opus/wav since there's no way to express those as targets. fine for the default (fallback on), just noting the edge.
- `audiophile` preset is now effectively 24-bit-only — that's a real behavior change for that preset (a 16-bit-only release would fail it with fallback off). intended? just flagging.
- the unified downloads response dropped the `[:limit]` slice / unbounded unverified-history load — minor payload/perf thing, not a break.

---

## bottom line

genuinely great feature and the architecture is solid. **#1 and #2 are hard blockers** (won't start / reintroduces a fixed bug) but both are small fixes. #3–#5 i mostly just need you to confirm are intended, and the rest are cleanup. get the two blockers + a green test run in and i think this is close.

no rush on the exam-season timeline — appreciate the work 🙏

nick2000713:
Thanks for the deep review and for actually running it — that's exactly the kind of pass this needed. All blockers + the flagged items are fixed; full `pytest` is green now (6602 passed, was 39 min / 26 fail / 17 err — the slowness was downstream of blocker #1). Pushed as 6 thematic commits.

  ### 🔴 Blockers
  - **#1 DB init crash** — moved `CREATE INDEX idx_lh_verification_status` to after the column-migration loop (next
  to `idx_lh_origin`). Confirmed a fresh `MusicDatabase()` initialises again. This was also the cause of three
  "hanging" test files — they were stuck on the aborted init, not real hangs.
  - **#2 #652 quarantine loop** — re-wired `_drop_quarantined_sources(results)` at the top of
  `filter_results_by_quality_preference`; the pinned test passes again.

   ### 🟠 Confirm-intended → fixed as you suggested
  - **#3** Agreed it shouldn't change the default flow. `search_with_fallback` (the priority path) is
  quality-agnostic again — first source with tracks wins, byte-for-byte. Cross-source ranking stays in
  `best_quality` mode + the import guard + the scanner.
  - **#4** Fixed `matches_target`: an unknown-spec FLAC (no sr/bd) now fails a strict hi-res target and falls to the
  plain-flac bucket, so it can't outrank a real 16/44 FLAC.
  - **#5** Good catch. Added a migration in `_migrate_v2_to_v3`: if a user had
  `tidal_download`/`qobuz`/`hifi_download.quality` on `hires`/`hires_max`, it seeds the 24-bit FLAC ladder at the
  top of `ranked_targets` (only when the profile isn't already 24-bit, so no duplication).

   ###  🟡 Nits
  - `test_quality_upgrade` red suite — rewired to the new `targets_from_profile` + `quality_meets_profile` API.
  While doing it I found a real bug: there were two `_fix_quality_upgrade` methods and the legacy one shadowed the
  findings-based one, so applying a finding failed live with "No title/artist". Merged into one handler
  (`matched_track_data` → wishlist, plus `_fix_action='delete'`).
  - `.aiff` now probed with `mutagen.aiff.AIFF` instead of `WAVE`.

  ## Left as-is (intended / separate)
  - `audiophile` being 24-bit-only is intended — that's the "24-bit only in one click" preset.

PR #936, opened 2026-06-27
Fix fragmented album rows in Album Completeness

ragnarlotus:
## Summary

- group local album rows that represent the same pinned canonical edition
- validate sibling-row tracks against the canonical tracklist before combining them
- emit one completeness finding per canonical edition instead of duplicate or contradictory findings
- preserve the canonical-edition no-fallback behavior introduced in #931

## Problem

A single physical album can be split across multiple local `albums` rows. Album Completeness currently evaluates those rows independently, so each row can appear incomplete even when their validated tracks collectively represent the same canonical edition.

Blindly merging rows by a shared source ID would also be unsafe: stale or incorrect IDs could combine unrelated albums and hide genuinely missing tracks.

Captura de pantalla 2026-06-24 175111
Captura de pantalla 2026-06-24 175120

## Fix

The job now builds logical work items before evaluating completeness:

- select a canonical anchor row for each candidate group
- only consider rows from the same artist that resolve to the same canonical source and album ID
- require strict local-track evidence before accepting a sibling row into the group
- match tracks one-to-one using MusicBrainz recording ID when available, then disc/track position, normalized title, and duration tolerance
- keep the anchor's observed count authoritative and add only validated canonical slots contributed by sibling rows
- keep rows independent when the shared ID is not supported by matching track evidence
- include related album IDs and validated local-track details in the resulting finding

This avoids false incomplete findings without allowing unrelated rows to be fused merely because they share an external ID.

## Testing

- Python syntax validation passed for the modified production and test files
- `14 passed` across:
  - `tests/test_album_completeness_job.py`
  - `tests/test_album_completeness_fragmented_rows.py`

- full suite: `6801 passed, 3 skipped`
- `git diff --check` passed against the current `dev` branch

### DIFF
--- a/core/repair_jobs/album_completeness.py
+++ b/core/repair_jobs/album_completeness.py
@@ -1,5 +1,10 @@
 """Album Completeness Checker Job — finds albums missing tracks."""
 
+import re
+import unicodedata
+from collections import defaultdict
+from difflib import SequenceMatcher
+
 from core.metadata_service import (
     get_album_tracks_for_source,
     get_primary_source,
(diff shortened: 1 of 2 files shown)


Nezreka:
this is fantastic work, thank you. really appreciate the care that went into it — the writeup, the tests, all of it.

and yeah, this solves a real problem. fragmented album rows (same physical album split across multiple `albums` rows) were getting evaluated independently, so each row looked incomplete on its own and you'd get duplicate/contradictory "incomplete" findings for what's actually one full edition. grouping the validated fragments into one logical album and emitting a single finding is exactly the right fix — and i like that you group by unambiguous id AND validate at the track level instead of blindly merging on a shared id. that's the safe way to do it.

found one small edge-case bug while reviewing: when two rows are pinned to the same canonical edition and one fails the strict fragment match, that excluded row ends up reporting the whole tracklist as missing (even the tracks it actually owns). minor and the download side self-corrects, so nothing scary — i'll patch it as a follow-up right after merge so i don't hold this up.

merging it. thanks again, this is genuinely helpful 🙏

PR #943, opened 2026-06-28
UI/settings page cleanup

nick2000713:
## Summary

  This PR cleans up and reorganizes the Settings page so it is easier to scan, less cramped, and grouped more
  naturally by task.

  It also fixes the saved UI appearance being applied too late on startup, which caused the app to briefly flash the
  default green accent before switching to the saved theme.

  ## What changed

  - Reorganized Settings into clearer tabs:
    - Connections
    - Downloads
    - Quality
    - Library
    - Appearance
    - Advanced
    - Logs

  - Converted larger Settings areas into collapsible tiles so the page is less overwhelming.

  - Moved related controls together without removing existing settings:
    - Source settings stay under Downloads
    - Quality profile and import quality controls moved under Quality
    - Folder paths moved under Library / Paths & Organization
    - Retry logic is its own Downloads tile
    - Indexer/downloader setup only appears when Torrent or Usenet is relevant

  - Improved the Settings help text pattern with consistent collapsible info panels.

  - Improved Hybrid source setup:
    - drag ordering
    - per-source config opening inline
    - source status dots
    - “Test all sources” button

  - Removed some dead/legacy Settings CSS.

  - Fixed saved appearance startup:
    - saved accent is rendered server-side before CSS paints
    - app also bootstraps saved appearance on load
    - this prevents the green flash before the saved theme appears

  ## Verification

  I checked that existing Settings controls were not lost:

  - No existing Settings IDs from `main` were removed.
  - No existing static `settings.js` ID references were removed.
  - No duplicate IDs were introduced.
  - Existing save/load wiring is still present for moved controls like:
    - download path
    - transfer path
    - staging path
    - music videos path
    - playlist materialize settings
    - M3U base path

  Also ran:

  - `git diff --check`
  - `node --check webui/static/settings.js`
  - `node --check webui/static/init.js`

  ## Notes

  Please try it out and see how the new Settings flow feels in practice. If any grouping, wording, spacing, or
  interaction feels off, I’m happy to adjust it.


nick2000713:
Just a small opinion from my side: for new users trying this for the first time, the first impression matters a lot.

Since I personally experienced significant performance issues unless “Reduce visual effects” was enabled, I think it would be better if animations and heavier visual effects were disabled by default. In my opinion, the default setting should prioritize performance and responsiveness, especially for first-time users, and users who want the extra animations could still enable them manually.

Nezreka:
Oooo this is beautiful. And you're right, I'll make that the default. Merging now!

PR #950, opened 2026-06-29
fix spotify cache mismatch

HellRa1SeR:
### Summary of Changes: Spotify OAuth Token Cache Unification

This PR fixes Spotify OAuth completing successfully but failing post-auth validation in Docker. The root cause is a split token store: 
 - OAuth callbacks wrote tokens to the legacy file cache (`config/.spotify_cache`) while `SpotifyClient` reads from the database-backed `DatabaseTokenCache` (`spotify.token_info`). 
 - After token exchange, validation probed with a stale or missing DB token
 - spotipy then attempted interactive OAuth on port 8888, which collided with SoulSync's dedicated callback server → `[Errno 98] Address already in use` and the warning *"token exchange succeeded but authentication validation failed"*.

This is a follow-up to the spotify database token migration in #812 `603b7a2a` (wolf39us spotify daily docker de-auth fixed).

**Key Changes:**

1. **Shared auth manager factory in `core/spotify_client.py`:**
   * Introduced `build_spotify_oauth_auth_manager()`. Builds `SpotifyOAuth` with `DatabaseTokenCache`, `normalize_spotify_oauth_config`, and `SPOTIFY_OAUTH_SCOPE` — the same stack `SpotifyClient._setup_client()` uses.
   * Refactored `_setup_client()` to call the factory instead of duplicating `SpotifyOAuth` construction.

2. **OAuth callback paths in `web_server.py`:**
   * **Flask `/callback` (port 8008 / reverse-proxy path):** Replaced `cache_path='config/.spotify_cache'` with `build_spotify_oauth_auth_manager()`.
   * **Dedicated callback server (port 8888 / Docker local path):** Same change.
   * Invalidate the auth cache on the global `spotify_client` immediately after token exchange, before `is_spotify_authenticated()`, so validation does not reuse a cached `False`.

3. **New unit tests in `tests/test_spotify_client.py`:**
   * `build_spotify_oauth_auth_manager` wires a `DatabaseTokenCache` (not `cache_path`).
   * Returns `None` when credentials are missing.

**Out of scope (unchanged):**

* Per-profile Spotify OAuth still uses per-profile file caches (`config/.spotify_cache_profile_{id}`) — that path is internally consistent with `get_spotify_client_for_profile()` in the metadata registry.

**Reasoning:**

`DatabaseTokenCache` was added so tokens survive Docker container recreation alongside other settings. The OAuth callback handlers were never updated to use it, so every successful handshake wrote to a file the runtime client did not read (or lost behind a stale DB entry). Unifying on one cache store removes the validation failure and prevents spotipy from binding port 8888 for interactive auth inside Docker.

**Test Plan:**

* **Unit tests:** `pytest tests/test_spotify_client.py tests/test_spotify_token_cache.py`
* **Manual verification (Docker):**
  1. `curl -s -X POST http://localhost:8008/api/spotify/disconnect` (or Disconnect in Settings)
  2. Authenticate via Settings → Connections (redirect URI `http://127.0.0.1:8888/callback` with port 8888 mapped)
  3. Confirm logs show *Spotify Auth Complete* (no `EADDRINUSE`, no *authentication validation failed*)
  4. Confirm Spotify shows connected in the UI and API calls succeed
* **Reverse-proxy path:** Repeat with redirect URI `https://yourdomain.com/callback` (Flask `/callback` on 8008)

### DIFF
--- a/core/spotify_client.py
+++ b/core/spotify_client.py
@@ -53,6 +53,36 @@ def normalize_spotify_oauth_config(config: Optional[Dict[str, Any]]) -> Dict[str
     return normalized
 
 
+def build_spotify_oauth_auth_manager(
+    config_manager=None,
+    *,
+    config: Optional[Dict[str, Any]] = None,
+) -> Optional[SpotifyOAuth]:
+    """Build SpotifyOAuth with the database-backed token cache.
+
+    OAuth callbacks and SpotifyClient must share the same cache store.
+    Callbacks that wrote to ``config/.spotify_cache`` while the client read
+    ``spotify.token_info`` from the DB left validation probing with a stale or
+    missing token; spotipy then tried interactive OAuth on port 8888 →
+    EADDRINUSE inside Docker.
+    """
+    from config.settings import config_manager as default_cm
+    from core.spotify_token_cache import DatabaseTokenCache
+
+    cm = config_manager or default_cm
+    cfg = normalize_spotify_oauth_config(config or cm.get_spotify_config())
+    if not cfg.get("client_id") or not cfg.get("client_secret"):
+        return None
+
+    return SpotifyOAuth(
+        client_id=cfg["client_id"],
+        client_secret=cfg["client_secret"],
+        redirect_uri=cfg.get("redirect_uri", "http://127.0.0.1:8888/callback"),
+        scope=SPOTIFY_OAUTH_SCOPE,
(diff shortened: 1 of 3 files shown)


Nezreka:
I actually committed this same fix earlier this morning :) Thank you though.

HellRa1SeR:
thanks! as long as its addressed. appreciate it. 
I love your app.

PR #962, opened 2026-06-30
feature jiosaavn: add album search and fix pagination for discography

HellRa1SeR:
## Summary

- **JioSaavn artist / album / track detail** — `/api/artist-detail`, `/api/spotify/album`, and `/api/spotify/track` now handle `source=jiosaavn` instead of 404ing on JioSaavn IDs returned from search. Adds `get_artist`, `get_artist_albums`, and `get_album_tracks` on `JioSaavnClient`, wires source-only artist detail, and registers `jiosaavn` in `SOURCE_ID_FIELD` + `source_ids` with DB `jiosaavn_id` columns on artists/albums/tracks.
- **Experimental provider gating** — Introduces `EXPERIMENTAL_SOURCES` + shared helpers (`is_source_enabled`, `experimental_source_rejected`, `experimental_status`) in `core/metadata/registry.py`. Backend routes (search fan-out, quick-switch, config-status, `/status`, settings save) and the search UI (`shared-helpers.js`) all respect per-provider flags — JioSaavn is hidden and rejected when `experimental.jiosaavn_enabled` is off.
- **Artist discography via album search** — `get_artist_albums` uses `/api/search/albums?query=<artist name>` (up to `limit` 200) instead of `/api/artists/{id}/albums`, which is hard-capped at 10 releases per page regardless of `limit`. Falls back to the first artist-albums page when the name cannot be resolved. Documented in the client module doc.
- **Singles** — No singles section for JioSaavn (upstream has no release-type singles; SoulSync does not synthesize one from the songs feed). Albums-only on artist detail; individual tracks remain searchable.

## Notes

- Album search discography is text-matched — soundtracks and compilations may appear alongside core artist releases.
- The upstream search API often returns fewer rows than the requested `limit` in a single response (~40 for prolific artists).
- Future experimental providers only need a registry entry + settings checkbox; gating and UI hiding are handled generically.

## Test plan

- [x] Enable **Settings → Advanced → Experimental → JioSaavn**; confirm JioSaavn appears in Search source picker and Connections quick-switch.
- [x] Disable the flag; hard-refresh Search — JioSaavn icon should disappear and `source=jiosaavn` detail requests should return 503.
- [x] Search **Badshah** on JioSaavn → open artist **456863**; confirm artist page loads (no 404) with albums listed.
- [x] Open a JioSaavn album and track from search; confirm detail pages load with `source=jiosaavn`.
- [x] Compare album count on JioSaavn artist detail vs before (~40 vs 10 for Badshah).
- [x] Confirm Singles tab is empty for JioSaavn artists (expected).
- [x] Run `pytest tests/metadata/test_jiosaavn_client.py tests/metadata/test_metadata_registry.py tests/metadata/test_metadata_album_tracks.py tests/search/test_search_orchestrator.py tests/metadata/test_artist_source_lookup.py tests/test_source_ids_registry.py tests/test_artist_full_detail_source_id.py`.


Nezreka:
I really appreciate how much work you are putting into the project. 🙏 Will review this shortly

core/metadata/album_tracks.py
Nezreka, line 1: hey, thanks for this one — the jiosaavn artist/album/track detail, the generic experimental-source gating, and the discography-cap fix are all clean and well-tested. one blocker before i can merge though, and it's a silent one so worth spelling out.

`get_artist_albums_for_source` (core/metadata/album_tracks.py:483) adds `'artist_name': artist_name` to the kwargs for **every** source, then passes it to `client.get_artist_albums(**kwargs)` at line 489. but only the jiosaavn client takes that arg. spotify's `get_artist_albums` (spotify_client.py:2038) and itunes' (itunes_client.py:1052) have no `artist_name` param and no `**kwargs`, so the call raises `TypeError: unexpected keyword argument 'artist_name'`. that then gets swallowed by the `except Exception: return None` at line 519 — so `get_artist_discography` just gets `[]` back and **artist discography comes up empty for spotify/itunes users, with no error shown.** since those are the primary providers for basically everyone, i can't pull it as-is.

the 6 failing tests in `tests/metadata/test_metadata_discography.py` are catching exactly this — they assert spotify is called *without* `artist_name`. they're correct, please don't update them; they should go green once the code's fixed.

fix is basically one line — gate `artist_name` to the client that actually accepts it:

```python
kwargs = {'album_type': album_type, 'limit': limit}
if source == 'jiosaavn':
    kwargs['artist_name'] = artist_name
if source == 'spotify':
    kwargs['allow_fallback'] = False
    kwargs['skip_cache'] = skip_cache
    kwargs['max_pages'] = max_pages
```

(or add `artist_name=None`/`**kwargs` to the other providers' signatures, but the gate is cleaner.)

everything else looks good — sort that one out and i'll merge. thanks again 🙏

HellRa1SeR: thank you for your review.
Absolutely agree - most users are on spotify/itunes and that cannot be broken.
This has been fixed. - appreciate it!

Issue #501, opened 2026-05-05
Enrichment and Reorganize should use manually matched ID


Tacobell444:
The Issue: Currently, if you manually match an album and then try and Enrich from that source or Reorganize the file Soulsync will rematch with whatever it thinks is the best match, re-matching to an incorrect album and enriching the wrong data or reverting to "Not Found"

What Should Happen: When a services is already matched, "Enrich" and "Reorganize" should use the matched ID instead of redundantly performing a new match.

Nezreka:
Great catch, I have the fix for this live on dev branch. Will push main soon

PR #968, opened 2026-07-02
feat: add Bandcamp as an experimental Discover search + enrichment source

bluejorts:
## Summary
- Bandcamp has no general-purpose public API (the real `bandcamp.com/developer` OAuth API is gated to label/merch-partner accounts), so this uses two endpoints Bandcamp itself serves unauthenticated: the public JSON autocomplete search API and each release page's embedded schema.org JSON-LD block. Both were verified against live Bandcamp pages during development — `bandcamp.com/search` (the HTML results page) is now behind a JS client-challenge for non-browser clients, so that page is intentionally avoided.
- Adds `core/bandcamp_client.py` (search + release metadata) and `core/bandcamp_worker.py` (background enrichment, covering both albums and tracks — albums are Bandcamp's primary unit, since a release's JSON-LD carries the full tracklist plus tags/label/credits in one fetch).
- Wires enrichment into `core/metadata/source.py`: tags/label/URL get embedded as real ID3/Vorbis/MP4 tags, with an album-search fallback when track search doesn't confidently match.
- Wires Discover search via the existing experimental-source opt-in mechanism (`core.metadata.registry.EXPERIMENTAL_SOURCES`) — off by default, same pattern as JioSaavn.
- Adds a Settings toggle, a dashboard "Manage Enrichment Workers" bubble, and Discover search-tab styling.
- DB migration adds `bandcamp_*` columns to `albums` and `tracks`.

## Follow-up fixes from live testing

The initial implementation looked complete on paper but missed several places this codebase hand-registers each metadata source individually — surfaced one at a time by actually importing/enriching/browsing real Bandcamp artists end to end:

**Import pipeline**
- Bandcamp was reachable in Discover/Enhanced Search but never in *manual import matching* (`core/imports/staging.py`, `core/metadata/album_tracks.py`) — a separate priority chain with no per-request source override.
- Artist name never reached the backend for URL-only sources like Bandcamp (routing assumed every source supported ID-only lookup); threaded a `?name=` param end-to-end.
- Opening an album from the artist-detail discography, or a Bandcamp artist/album from search, both 404'd — more of the same "assumes an ID-lookup API" gap.
- **General (not Bandcamp-specific) hardening**: imports now refuse to run unless the active media server is either standalone (`soulsync`) or actually connected, returning a clear 503 with a Settings link instead of silently copying files onto disk without adding them to the library DB.

**Library view**
- Added Bandcamp to the artist Enrichment Coverage panel, the derived per-source badge next to the artist name (no `artists.bandcamp_id` column by design — derived from any owned album/track's release URL), and the enhanced library view's artist/album/track match chips, ID badges, and Enrich dropdowns (album/track only — Bandcamp has no artist-level enrichment).
- Manual-match search for Bandcamp returned zero results: the dispatcher was calling the confidence-gated `search_album`/`search_track` convenience methods (tuned for the unattended auto-enrichment worker) instead of the raw multi-result search a human picking from a list actually needs. Also narrows per-track manual search by including the album title alongside the track title.
- **Real data-integrity bug**: `BandcampWorker`'s "already matched, re-fetch metadata" fast path unconditionally nulled `bandcamp_id` on every re-enrichment pass (even though `bandcamp_url`/`bandcamp_match_status` stayed correct), which silently broke both the new per-track match chip and the enrichment-coverage percentage after the first re-run. Fixed with `COALESCE` so a missing id preserves the existing one.
- **Missing artist/album art, two distinct causes**: (1) the shared image cache (`core/image_cache.py`) hardcoded a `Referer: https://www.deezer.com/` header on every artwork fetch — a leftover from when it only served Deezer covers — which Bandcamp's hotlink-protected CDN silently rejected; (2) separately, `search_album`/`search_track` preferred the autocomplete search index's cached thumbnail over the release page's own embedded image, and that cached thumbnail can point at a since-removed CDN size variant (confirmed via a live 404 against an already-imported artist).

## Follow-up fixes from @Nezreka's review

A second sweep against how jiosaavn/deezer are wired everywhere, catching the lists a new source still had to be hand-added to:

- **Runs when off** — gated the auto-import hook, the enrichment `source_order` strip, and single-item re-enrich on `is_source_enabled('bandcamp')`.
- **Reset couldn't clear a match** — registered bandcamp in `_CORE_ID_COLUMNS` (canonical column = `bandcamp_url`, since Bandcamp has no id→page lookup) so reset nulls the match on albums *and* tracks; coverage now counts `bandcamp_url` (fixes matched-badge-but-0%).
- **ratingKey migration** — carries the `bandcamp_*` album columns so a rekey no longer drops the match.
- **Worker lifecycle/status parity** — paused during DB scans, `.stop()`-joined on shutdown, added to `_get_enrichment_status`, `SERVICE_CONFIG_REGISTRY`, the api-call rate-monitor maps, and the priority-override hook.
- **Enrichment worker brought in line with peers** — routes through the shared `honor_stored_match` id-refresh path (#501, no re-search / no manual-match stomp) and backfills the shared album cols (`label`/`release_date`/`genres`/`api_track_count`).
- **Rate limiter** no longer sleeps while holding the global lock.
- **Frontend parity** — worker orb, dashboard enrichment grid, stats coverage row; removed Bandcamp from the *artist-level* enrich menu (album/track only); scoped the album-title-in-query to Bandcamp manual matches.

## Test plan
- 7520 backend tests pass (`python -m pytest tests/`), 96 frontend tests pass (`npx vitest run` in `webui/`).
- New regression tests added throughout for each fix above: the import media-server gate, the image-cache Referer bug, the Bandcamp client's image-preference fix, the worker's `bandcamp_id` COALESCE fix, and the manual-match search dispatch.
- `node --check` / `ast.parse` clean on all changed JS/Python files.
- Manually verified against live Bandcamp (search, release-page scrape, and the specific broken-thumbnail-URL repro) during debugging — not just mocked.


HellRa1SeR:
Love this! more sources is always good. I will check this out when its ready 🔥

bluejorts:
@HellRa1SeR I'm not in love with the way this has to pass more parameters through than most other sources, but it's a limitation of Bandcamp's lack of official API. They don't have any underlying artist IDs available, so Artist Name and Release Name are what we have to go off of. Polishing it up and will submit later today or this weekend. It's working locally for me, for a few niche single-release artists I have on Bandcamp that don't exist anywhere else.

I'd also like to make a second PR after this is merged to enable full import of a Bandcamp collection given a user's Bandcamp URL. Unclear if that should also make one way Sync jobs possible or not. It would be nice to have music purchased on Bandcamp automatically end up in SoulSync. I'm interested in enabling that kind of "purchase -> download" pipeline as it encourages more use of direct-artist-payment services rather than streaming.

HellRa1SeR:
> I'm not in love with the way this has to pass more parameters through than most other sources, but it's a limitation of Bandcamp's lack of official API.

@bluejorts Oh yes, I completely understand and agree. I am working on two providers that dont have official APIs, but trying to reverse engineer their apis. So, it is pretty inconsistent in how we write the client and the fetch apis which might be complex and convoluted and not as straightforward as deezer/spotify.
Thank you for the work - more meta information/providers is always good.

I've used MB Picard, Beets and a barrage of other suites, and nothing is fail proof. Hopefully, SoulSync is that product.

Nezreka:
first off, this is awesome — adding a whole new source is a big lift and you clearly put real care into it. when it's enabled it mostly just works, which is impressive given how many places a source has to touch. i gave it a full pass and then went through it like it's another metadata source, checking bandcamp against how jiosaavn/deezer/itunes are wired everywhere. good news is it's not structural, it's just that a new source has to be hand-added to a ton of lists and a few got missed. here's what i found, roughly in priority order 🙏

**worth fixing before merge**
- it's running even when it's off. bandcamp doesn't get stripped from the source order when disabled (jiosaavn does), `_process_bandcamp_source` doesn't check `is_source_enabled`, and the single-item re-enrich (`web_server.py` ~12980) doesn't gate on it either — so both auto-import and manual re-enrich hit bandcamp.com even with the toggle off. probably the main one to sort.
- a bad match can't be cleared. bandcamp's not in `_CORE_ID_COLUMNS` (`core/source_ids.py`), so the reset path never nulls `bandcamp_url` and the worker just re-confirms the old release. clear-match handles it but reset doesn't, so the two disagree.
- ratingKey migration loses it. `enrichment_cols` in `music_database.py` (~6567) carries jiosaavn but none of the `bandcamp_*` cols, so when a plex/jellyfin album's ratingKey changes the bandcamp match/url/tags/label get dropped and it re-scrapes.

**parity bits it got skipped on (all small)**
- not paused during db scans (`_pause_workers_for_scan`), so it keeps writing mid-scan — the lock contention that's meant to prevent
- not in `_get_enrichment_status` workers_info or the rate-monitor key map, so no worker badge in system status
- doesn't honor the manage-workers priority override (never calls `read_enrichment_priority`/`priority_pending_item`)
- not `.stop()`-joined on shutdown (it's a daemon thread, so it gets killed mid-write)
- `SERVICE_CONFIG_REGISTRY` is missing `bandcamp: {always: True}`, so it reads as "not configured"
- coverage counts `bandcamp_id`, but manual/url-only matches only write `bandcamp_url` (id stays null), so matched items show a badge but 0% in coverage
- frontend: no worker-orb def, and it's missing from the dashboard enrichment grid + the stats coverage list. also the artist enrich menu offers bandcamp even though there's no artist pass

**enrichment worker — mostly a heads-up on the standard the others follow**
right now it's basically "search → store the url → save tags/label in its own columns," and the other workers do a bit more. a few things that'd bring it in line:
- capture + store a stable id. the search box actually hands you real numeric item-ids, but they get stored inconsistently (sometimes the numeric id, sometimes the url, sometimes `{kind}-single`, null on the manual/url paths). so re-enrich ends up re-searching instead of a direct lookup, and `bandcamp_id` (which coverage counts) is unreliable. also wants to go in `_CORE_ID_COLUMNS`.
- lean on the `honor_stored_match` id-refresh path like deezer/qobuz/spotify/tidal/jiosaavn do (#501) — if it's already matched, refresh by id instead of re-searching, without stomping a manual match. right now it rolls its own url short-circuit.
- persist more of what it already pulls. i ran the parser and it's actually getting artwork, release_date, full tracklist + durations, label and tags — but the worker only saves tags/label into its own namespace and drops the rest, and doesn't feed the shared cols the peers write (`albums.label/genre/explicit/release_date` + `api_track_count`). so it lands a "matched" badge but doesn't add much to the album's real metadata yet. (the import-time tag/label embedding is great as-is — it's just the background enrichment that's light.)

**minor stuff**
- the rate limiter sleeps while holding the global lock, so clicking a bandcamp album can stall ~1s behind the worker — releasing the lock before the sleep would fix it
- the manual-match query change in `library.js` (~4505) hits all services, not just bandcamp — single tracks on spotify/deezer now get "album title track title" prefilled
- `image_cache.py` referer went from hardcoded-deezer to same-origin — might be worth a quick check that deezer art still caches

honestly none of this takes away from how solid the core is — it's mostly "new source needs adding to list X," which is easy to miss when there's no single source registry to add it to. 🙏

bluejorts:
I'll take a look at this feedback when I have some time this week! Would there be appetite for a possible refactor into a protocol based method where sources have a promise to provide certain data/features, and centralize a source registry? Thinking something along the lines of:

```
- bandcamp
  - album
    - import
    - enrich
    - search
  - track
    - import
    - enrich
    - search
- deezer
  - album
    - ...
  - track
    - ...
  - artist
    - ...
```

bluejorts:
Thank you for the incredibly thorough pass — you were right that it's "new source needs adding to list X" almost everywhere, and going through it against how jiosaavn/deezer are wired made each one easy to find. All of it is addressed now (pushed to `bluejorts:feature/bandcamp-source`). Rundown, in your priority order:

### Worth fixing before merge
- **Running when off** — gated all three paths: `_process_bandcamp_source` and the enrichment `source_order` strip now check `is_source_enabled('bandcamp')` (mirroring jiosaavn), and the single-item re-enrich dispatcher returns the "disabled" error like jiosaavn does.
- **Bad match can't be cleared** — registered bandcamp in `_CORE_ID_COLUMNS` and taught `build_reset_query` to clear it. One wrinkle worth calling out below (§1).
- **ratingKey migration** — added the six `bandcamp_*` columns to the album `enrichment_cols` copy list.

### Parity bits
- Paused in `_pause_workers_for_scan`/`_resume_workers_after_scan`; `.stop()`-joined in `_shutdown_runtime_components`; added to `_get_enrichment_status` (`workers_info` + `configured_checks` + the disabled-skip); `SERVICE_CONFIG_REGISTRY` now has `bandcamp: {always: True}`; added to `api_call_tracker` (`RATE_LIMITS`/`SERVICE_LABELS`/`SERVICE_ORDER`) and `debug_info.worker_names`.
- Coverage now counts `bandcamp_url` instead of `bandcamp_id` (see §1) — fixes the matched-badge-but-0% mismatch.
- Priority override: `_get_next_item` now calls `read_enrichment_priority`/`priority_pending_item` like the other workers (album/track only).
- Frontend: added the worker orb, the dashboard enrichment grid entry (gated by a new `isBandcampExperimentalEnabled()`), and the stats coverage row; **removed** bandcamp from the *artist-level* enrich menu (no artist pass) while leaving the album-level menu. Also scoped the `album title + track title` manual-match query to bandcamp only.

### Enrichment worker
- Now routes through the shared `honor_stored_match` (#501) fast path and persists `label` / `release_date` / `genres` (from tags) / `api_track_count` into the shared album columns, backfill-only.

### Two design decisions I'd like a sanity check on

**§1 — Bandcamp's stable id is the release URL, not the numeric item-id.** The autocomplete box does hand back numeric item-ids, but there's no `id → release page` endpoint we can hit — `get_release_metadata` takes a URL, and the release-page JSON-LD carries no id. So the URL is the only *resolvable* stable handle. I therefore made `bandcamp_url` the canonical column in `_CORE_ID_COLUMNS`, which is what `honor_stored_match` (refresh-by-URL, no re-search, no manual-match stomping), the reset path, and coverage all key on now — consistent with the existing `_backfill_match_status_for_existing_ids`, which already treats `bandcamp_url` as the id-signal. The numeric `bandcamp_id` is still captured and COALESCE-preserved as a supplementary key (and the reset nulls it too, on both albums and tracks). Net effect matches what you asked for — re-enrich does a direct fetch instead of re-searching — it's just keyed on the URL rather than the numeric id. Happy to revisit if you'd rather store/lookup on the numeric id somehow.

**§2 — I left `thumb_url` out of the shared-col writes.** The parser does pull artwork, but Bandcamp's bcbits CDN is hotlink-protected and the art already flows through `image_cache` with the same-origin referer fix. Writing a raw CDN URL into `albums.thumb_url` would render broken outside that proxy, so I persisted everything *except* artwork. Easy to add if you'd prefer it stored.

Tests: full backend suite green (7520 passed) and frontend green (96 vitest + the `tests/static` node suites), with new regression coverage for the gating, the reset clear, the ratingKey preservation, coverage-by-url, the priority override, `honor_stored_match` + shared-col persistence, the lock-free rate-limit wait, and the stats experimental gating.

PR #969, opened 2026-07-02
[torrent] match RuTracker album bundle files

ramonskie:
DRAFT:
this is a draft. as i did not test the whole worflow.
i just want to start a discussion. as rutracker is not working with lidarr either.
and i want to know if this is the right approach.

## Summary

Fix torrent/usenet album-bundle staging for RuTracker-style releases.

RuTracker music torrents often download files with release-style names instead of clean tags, e.g.:

- `Nosferatu_-_Beaver_Cleaver.flac`
- `01 - 4 Tune Fairytales - Take Me 2 Wonderland (Extended Mix).flac`

The album-bundle flow already downloads and stages these files correctly, but `try_staging_match()` compared raw filename stems against clean Spotify track titles. That caused staged album files to be missed, especially for compilations and artist-prefixed filenames.

## Changes

- Parse common release filename shapes during private album-bundle staging:
  - `Artist_-_Title`
  - `Artist - Title`
  - `NN - Artist - Title`
  - `NN - Title`
- Use parsed filename title/artist as additional match candidates.
- Use explicit track numbers as a guard/boost for private album-bundle matches.
- Skip stale staged-file cache entries when the file was already claimed and removed.
- Add integration coverage using real `TorrentDownloadPlugin.download_album_to_staging()` with mocked Prowlarr/torrent-client boundaries.
- Add RuTracker source examples as regression fixtures:
  - https://rutracker.org/forum/viewtopic.php?t=3503447
  - https://rutracker.org/forum/viewtopic.php?t=1599889

## Why

This fixes album downloads where the torrent completes successfully but tracks are not claimed from private staging due to filename/tag mismatch.

## Tests

- `python -m pytest tests/downloads/test_album_bundle_rutracker_integration.py tests/downloads/test_downloads_staging.py -q`
- `python -m ruff check core/downloads/staging.py tests/downloads/test_album_bundle_rutracker_integration.py tests/downloads/test_downloads_staging.py`
- `python -m pytest`

### DIFF
--- a/core/download_plugins/album_bundle.py
+++ b/core/download_plugins/album_bundle.py
@@ -101,6 +101,51 @@ def quality_score(title: str, quality_guess) -> int:
     return _QUALITY_SCORE.get(quality_guess(title) or '', 0)
 
 
+def album_search_queries(album_name: str, artist_name: str) -> list[str]:
+    """Return deduped Prowlarr query variants for album-bundle search.
+
+    Some indexers (notably RuTracker through Prowlarr) match release titles
+    better when the query mirrors tracker naming (``Artist - Album``), while
+    compilations are often listed under ``VA`` instead of ``Various Artists``.
+    Keep the ladder small: broad enough to find tracker-shaped titles, but not
+    so broad that we spam indexers or invite unrelated albums.
+    """
+    album = (album_name or '').strip()
+    artist = (artist_name or '').strip()
+    if not album and not artist:
+        return []
+
+    candidates: list[str] = []
+    if artist and album:
+        is_va = artist.lower() in {'various artists', 'various', 'va'}
+        candidates.extend([
+            f"{artist} {album}",
+            f"{artist} - {album}",
+        ])
+        if is_va:
+            candidates.extend([
+                f"VA {album}",
+                f"VA - {album}",
+                f"Various Artists - {album}",
+            ])
+        candidates.extend([
+            f"{album} {artist}",
+            album,
+        ])
+    else:
+        candidates.append(album or artist)
+
+    out: list[str] = []
+    seen: set[str] = set()
+    for candidate in candidates:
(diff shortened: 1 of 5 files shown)


ramonskie:
Torrent/usenet album-bundle downloads are already release-level: the selected torrent/NZB downloads the whole album/release. The per-track part happens only after download, during SoulSync’s existing import/post-processing handoff.

Current flow:

1. Pick one torrent/usenet release for the album.
2. Downloader fetches the whole release.
3. SoulSync copies all audio files into private album-bundle staging.
4. Existing per-track workers run for each expected album track.
5. Each worker calls `try_staging_match()` to claim the staged file that belongs to that track.
6. The matched file is post-processed/tagged/moved using that track’s metadata.

This PR does **not** make torrents download per-track. It fixes the current per-track staging handoff by teaching it to understand RuTracker-style release filenames like `Artist_-_Title` and `NN - Artist - Title`.

Longer term, the better architecture would be an album-level manifest matcher:

- build a manifest from all downloaded files
- compare the whole manifest once against the expected album tracklist
- assign file → track mappings globally
- let per-track post-processing consume those explicit assignments

That would better handle weak filenames, duplicate compilation tracks, durations, disc numbers, and partial/extra files. This PR intentionally avoids that larger redesign and keeps scope limited to making the existing post-download handoff work for common RuTracker album releases.

Zombiehamser:
Sorry, my question is a bit off-topic. However, I can’t help but ask: how do you use rutracker? Prowlarr can’t bypass Cloudflare protection, even with additional plugins.

ramonskie:
rutracker works perfectly fine with flaresolver.
here is some documentation https://github.com/ramonskie/groovearr/blob/main/docs/prowlarr-integration.md

but soulsync can never really work. as at its core it does per track searching.
i wrote my own app. that does make a distinction between full albums and per track.
and with fallback mechanisms. that perfectly integrates with rutracker

PR #974, opened 2026-07-02
quality-profiles: add named Quality Profiles (foundation for per-item quality rules)

nick2000713:
Hi, before you dig in....  I'm also working on a separate, bigger overhaul on my fork — moving the library toward a Lidarr-style Library Manager (opt-in library, monitoring, watchlist/wishlist, etc.). That work is still in progress and not ready to show yet.

This PR is **not** part of that...  it stands completely on its own. While working on the library stuff I kept running into the same wall: quality handling is one global setting that every workflow shares, and that doesn't scale once different parts of the app (or different items in the library) need different quality rules. So I pulled this out into its own branch and finished it independently, because it's useful right now regardless of what happens with the bigger library work later. Figured it's worth sending over rather than letting it rot on a branch.

## Summary

Replaces the single global quality setting with named, CRUD-able **Quality Profiles**. Quality targets, upgrade behavior, AcoustID strictness, downsampling, lossy-copy behavior, and import replacement rules can now be resolved per profile instead of always reading one global setting.

## Why

- prepare support for song-, artist-, album-, playlist-, or source-specific Quality Profiles later
- let Auto-Import use different quality requirements than normal downloads
- let the download/import pipeline resolve quality behavior per item
- support Lidarr-like "upgrade until target" behavior
- give the app-wide default a real, editable identity instead of a scattering of loose settings

## What This Enables Now

- Multiple named Quality Profiles can exist; one is the app-wide default.
- Auto-Import can be pointed at its own Quality Profile, independent of the default.
- Wishlist and library-track rows can carry a `quality_profile_id`; unset rows resolve to the default at read time.
- Import checks, candidate ordering, AcoustID strictness, downsample, lossy-copy, and replace-lower-quality all resolve through the assigned profile.
- Quality Check and Quality Upgrade Finder are profile-aware, including "upgrade until selected target" cutoff behavior, which now lives on the profile itself.
- A Manage view shows what's active for what (default vs. Auto-Import vs. per-item overrides), separately from previewing/editing a profile's settings.

## Groundwork For Later

This intentionally lays the foundation for deeper Lidarr-like behavior down the line:

- per-song / per-artist / per-album / per-playlist Quality Profiles
- profile-based repair jobs and profile-aware library scans
- different import/download rules depending on source or context

The architectural shift is that components read a *resolved profile* instead of a global setting, so they can become profile-dependent without another rewrite.

## Migration / Update Safety

Existing users are migrated automatically, once, on startup:

- old global quality settings are materialized into the (renamed) default Quality Profile
- existing Wishlist and library-track rows are backfilled with a profile pointer where resolvable
- the old `import.quality_filter_enabled = off` toggle is migrated into a dedicated, Auto-Import-scoped "accept anything" profile — Auto-Import gets the relaxed behavior it had before, without loosening the default profile used everywhere else
- deleted or missing profile references fall back to the default profile instead of breaking imports/downloads
- manual imports are unaffected and keep resolving through the app-wide default, matching pre-existing behavior

## Testing / Verification

- Unit/integration tests for schema creation and additive migrations, legacy-settings migration into profiles, Wishlist/library backfill, Auto-Import profile assignment, profile-aware import pipeline behavior, and Quality Check / Quality Upgrade Finder cutoff behavior.
- Full local pytest run green (7200+ tests).
- Migration verified against a real, pre-Quality-Profiles production database (copy and direct mount), confirming 100% backfill coverage with no data loss or errors.
- Docker/WebUI production build built and manually tested end-to-end (Settings, Manage modal, Auto-Import) before opening this PR.

This is not the final per-song/artist/album assignment system yet, but it's the necessary foundation so the rest of the app can become profile-aware cleanly.


Nezreka:
thanks for this — really solid PR. clean architecture (the single `load_profile_by_id` + cached resolution is the right call), thorough cleanup, good tests. no complaints on the shape.

one thing i'd fix before merge:

**`require_top_target` silently stops working after upgrade.** the bridge `if policy is None and settings.get("require_top_target")` never fires because `_quality_profile_row_to_dict` does `row["upgrade_policy"] or "acceptable"` (never None), and the migration reads preferences.quality_profile which never had that setting anyway. so anyone who had "upgrade until top quality" on silently drops to "flag below any target." fix is probably honoring it when policy is `acceptable`, not just None.

couple minor ones, not blockers:
- editing the default profile then saving settings can silently revert it — `sync_default_quality_profile_from_config` overwrites the row from config, but the update endpoint only writes the row, never config
- the relaxed auto-import profile gets INSERTed before the `auto_import.quality_profile_id` check, so it can leave an orphaned row nothing points at
- dropping `import.quality_filter_enabled` is a real behavior change for anyone who ran it off — looks intentional, just worth a release-notes line

nice work overall, mainly just want #1 handled so nobody loses upgrade behavior quietly 🙏

nick2000713:
Thanks, fixed and pushed.

PR #987, opened 2026-07-05
Prefer commercial sources for library discographies

ragnarlotus:
## Summary

- use iTunes as the primary source for library artist discographies
- fall back to Deezer only when iTunes is unavailable
- keep locally owned releases that are missing from the commercial catalogue
- preserve each release's provider source for completion checks, track lookup, and batch wishlist actions
- integrate with the existing `secondary_types` filtering already present in current `dev`

## Problem

Library artist pages currently depend on the globally configured metadata source. With MusicBrainz selected, the discography can become cluttered with bootlegs, broadcasts, live recordings, deluxe editions, and other release groups that are not useful as a commercial catalogue.

Switching the global metadata source is not desirable because MusicBrainz remains useful for matching and enrichment. Library discographies therefore need their own source policy.

## Implementation

- add a dedicated library-discography service
- query iTunes first and Deezer second, without falling back to MusicBrainz for the commercial catalogue
- merge omitted owned releases conservatively, avoiding duplicate editions and same-named singles
- retain the provider source on every release card
- keep explicit source overrides unchanged
- use per-release sources for completion checks, track lookup, and batch wishlist actions
- preserve compatibility with the current MusicBrainz `secondary_types` filtering and canonical-completion logic in `dev`

## Validation

Validated against current `dev` at `9040201fa4f158d23664fc9c5a0de6479664d630` (SoulSync 2.8.9):

- Ruff passed for the modified Python files and for the full repository
- Python compilation passed
- JavaScript syntax validation passed
- focused regression suite: 146 passed
- full Python suite: 7,646 passed, 2 deselected
- WebUI checks passed with 0 errors and 0 warnings
- WebUI build passed
- WebUI test suite: 96 passed
- canonical-completion regression tests: 3 passed
- deployed the candidate files to a SoulSync 2.8.9 instance and verified the service remained healthy
- live API check for The Smashing Pumpkins returned an iTunes-led catalogue with owned MusicBrainz releases merged back in
- manual UI verification confirmed the commercial discography is shown correctly
- after rerunning canonical resolution following a full import, owned/completion states were restored correctly

## Scope

The pull request changes 4 files and is based directly on current `dev`. Existing canonical completion fixes and current release-type filtering remain intact.

### DIFF
--- a/core/metadata/library_discography.py
+++ b/core/metadata/library_discography.py
@@ -1,0 +1,345 @@
+"""Library artist discography policy.
+
+The library view intentionally uses commercial catalogues instead of the
+configured metadata primary source.  iTunes is preferred, Deezer is the only
+external fallback, and locally owned releases omitted by both catalogues are
+merged back into the result.
+"""
+
+from __future__ import annotations
+
+import logging
+import re
+import unicodedata
+from difflib import SequenceMatcher
+from typing import Any, Callable, Dict, Iterable, List, Mapping, Optional, Sequence
+
+from core.metadata.lookup import MetadataLookupOptions
+
+logger = logging.getLogger(__name__)
+
+COMMERCIAL_DISCOGRAPHY_SOURCES: Sequence[str] = ("itunes", "deezer")
+
+_SOURCE_ID_FIELDS = {
+    "itunes": "itunes_album_id",
+    "deezer": "deezer_id",
+    "musicbrainz": "musicbrainz_release_id",
+    "spotify": "spotify_album_id",
+}
+
+_VARIANT_TERMS = (
+    "anniversary",
+    "clean",
+    "collector",
+    "deluxe",
+    "edition",
+    "expanded",
+    "explicit",
+    "legacy",
+    "mono",
+    "remaster",
+    "remastered",
+    "reissue",
+    "redux",
+    "special",
+    "stereo",
+    "super deluxe",
+    "version",
+)
+_VARIANT_GROUP_RE = re.compile(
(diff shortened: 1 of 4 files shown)


ragnarlotus:
Rebased and rebuilt this PR on current `dev` / SoulSync 2.8.9 (`9040201fa4f158d23664fc9c5a0de6479664d630`).

Validation completed:

- 7,646 Python tests passed, 2 deselected
- 146 focused regression tests passed
- 96 WebUI tests passed
- Ruff, Python compilation, JavaScript syntax, WebUI checks, and WebUI build passed
- canonical-completion regression tests passed
- deployed and smoke-tested on a live SoulSync 2.8.9 instance
- verified The Smashing Pumpkins now uses an iTunes-led commercial discography with owned MusicBrainz releases merged back in
- manually verified the UI and completion states after rerunning canonical resolution following a full import

The head branch now contains a single commit based directly on current `dev`: `c8268b5725ffed359940c09060c14537efd20ecb`.

ragnarlotus:
After testing this approach, I think there is a cleaner and more flexible design for this feature, so I am closing this pull request rather than extending the current implementation.

The current PR hard-codes a commercial-discography policy and relies on title heuristics to merge local releases. A better solution is to add a separate configurable **Library discography source** alongside the existing primary metadata source. That would let users keep MusicBrainz (or another provider) for general metadata and enrichment, while choosing a more conservative provider such as iTunes specifically for the album/EP/single catalogue shown on library artist pages.

I plan to rework this as a new pull request with that configuration model and a more deterministic merge strategy based on provider IDs and canonical resolution instead of fuzzy title matching.

Issue #765, opened 2026-06-02
Conflict between Library Reorganizer and Track Number Repair due to different album versions


CubeComming:
When I run the 'Library Reorganizer', it prompts me to change a track number from '3' to '04', which is correct. However, when I run the 'Track Number Repair' tool, it fails to update the track number in the metadata.

The root of the problem is that online databases list different versions of the same album. For example, with Imagine Dragons' album Evolve, streaming services like Spotify place the track 'Believer' at position 4, while MusicBrainz has it listed as track 3. Switching to the correct database version naturally brings along other tag changes that need to be updated as well, not just the track number. Because the tools aren't synced, they get confused and constantly contradict each other.

Ideally, they should work together—one moving and renaming the files, and the other updating the track numbers and all corresponding metadata. Alternatively, a single tool that handles file reorganization, renaming, and tag updates all at once would be much better.

Nezreka:
I have this resolved on dev, will update main soon.

Issue #758, opened 2026-06-01
[Enhancement] Lock enrichment IDs when manually matched


Tacobell444:
I have multiple albums where the auto enrichment really wants to match them to the deluxe edition. I go in and manually match it to the regular edition. The next enrichment cycle it gets matched back to the deluxe edition, reporting missing songs or renumbering songs in a different order than the standard edition. 

When you manually match an album, it should lock and use the manually matched ID for future enrichments. 

I know you fixed a similar issue with the "enrich" button previously, maybe this is similar.

Tacobell444:
@Nezreka I saw #777 come through, seems closely related to this. Does that update include 'Pinning' a manually matched ID? If so, this can be closed.

Nezreka:
Now it does :D Newest update will have this resolved

PR #1022, opened 2026-07-14
fix(repair): apply canonical_version findings (add _fix_canonical_version)

sam-coodu:
## Summary

Canonical-version findings could not be applied. Clicking Fix (or Fix All) always returned `No fix available for finding type: canonical_version` and pinned nothing.

## Problem

The **Resolve Canonical Album Versions** job (#765) creates `canonical_version` findings during its dry run, and the UI offers a Fix action on them, but applying one is a no-op error.

## Root cause

`RepairWorker._execute_fix` has no handler entry for `canonical_version`, so it falls through to the generic `No fix available for finding type: ...` branch. The scan side exists; the apply side was never wired up.

## Solution

Add `_fix_canonical_version`. It pins the resolver's chosen release (`source`, release id and `score`, taken straight from the finding) via `db.set_album_canonical` as an AUTO pin (`locked=False`). This is the per-finding equivalent of running the job with dry-run OFF, matching `resolve_and_store_canonical_for_album` and the Reorganizer, so a later resolve can still self-heal it.

It deliberately does not create a manual/locked pin (that is the album-view edition override, #758) and will not clobber an existing locked pin. Malformed findings are rejected rather than written as a null pin.

## Testing

New `tests/test_repair_worker_canonical_version.py` covers the apply path, the dispatch fix (no more "No fix available"), the malformed-finding guard, and the manual-lock guard.

Locally: `ruff check .` clean, `py_compile` clean, and the 172 existing canonical/repair-worker tests plus the 4 new ones pass.

### DIFF
--- a/core/repair_worker.py
+++ b/core/repair_worker.py
@@ -1029,11 +1029,50 @@ class RepairWorker:
             'library_retag': self._fix_library_retag,
             'short_preview_track': self._fix_short_preview_track,
             'corrupt_audio': self._fix_corrupt_audio,
+            'canonical_version': self._fix_canonical_version,
         }
         handler = handlers.get(finding_type)
         if not handler:
             return {'success': False, 'error': f'No fix available for finding type: {finding_type}'}
         return handler(entity_type, entity_id, file_path, details)
+
+    def _fix_canonical_version(self, entity_type, entity_id, file_path, details):
+        """Apply a canonical-version finding — pin the release the resolver chose
+        (source, release id and score, straight from the finding) onto the album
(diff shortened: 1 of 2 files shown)


sam-coodu:
CI note: `build-and-test.yml` triggers on `push` (non-main/dev branches) rather than `pull_request`, so a fork PR doesn't kick off checks here automatically. I ran the pipeline on my fork against this exact branch and it's green:

https://github.com/sam-coodu/SoulSync/actions/runs/29302801937

Both jobs pass: `sanity-check` (ruff + compileall + full `python -m pytest`, including the 4 new canonical-version tests) and `webui` (lint + build + test).

PR #1032, opened 2026-07-15
Fix: surface discography provider access failures

ragnarlotus:
## Summary

Artist discography currently treats two different outcomes as the same state:

- a valid provider response with no releases;
- a provider access failure such as timeout, DNS/TLS/network error, refused connection, invalid HTTP response, or upstream outage.

That causes technical failures to be silently converted into an empty catalogue and allows fallback to continue, which can hide the real problem and return data from another provider.

This PR introduces a strict three-state contract for artist-discography lookups:

- `results`: the provider completed and returned releases;
- `empty`: the provider completed successfully but returned no releases;
- `error`: communication with the provider could not be completed.

A valid empty response may continue through the configured fallback chain. A provider-access failure stops fallback immediately and is returned to the web layer with its HTTP status and sanitized message.

A provider may retry equivalent mirrors or requests internally. When that same provider eventually returns releases, the recovered result wins; a captured technical failure only explains an otherwise empty result.

> A valid query with no results and a communication failure do not represent the same state and should not produce the same behaviour.

## Source-authoritative direction

This PR is intentionally an intermediate compatibility layer toward source-authoritative lookups.

When a user explicitly selects a metadata source, that source is intended to become the source of truth. Internal retries within that provider may recover, but an access failure from the selected provider must not be silently replaced with results from another provider.

A follow-up can move the `results` / `empty` / `error` distinction into the provider clients themselves, allowing the request instrumentation layer introduced here to be simplified substantially or removed.

## Scope

The strict behaviour is intentionally limited to the two user-facing artist-discography web paths:

- artist detail;
- download discography.

The legacy `core.metadata_service` facade remains lenient. Existing behaviour for background repair jobs, search, import, enrichment, and unrelated metadata flows is unchanged.

## Implementation

- Adds request-local provider instrumentation that preserves transport, HTTP, parsing, and wrapped client failures without mutating shared registry clients.
- Adds strict artist-discography orchestration with explicit `results` / `empty` / `error` states.
- Preserves valid `400`, `404`, and `410` responses as empty only during explicit artist-ID resolution.
- Keeps Spotify no-auth handling inside the Spotify catalogue while preventing numeric IDs from triggering cross-provider fallback.
- Treats a recovered same-provider call returning releases as success, even if an earlier internal attempt failed.
- Routes only artist-detail and download-discography endpoints through the strict facade.
- Keeps source-only artists as hard failures when their provider is inaccessible.
- Keeps owned artist pages available on HTTP 200 with owned releases plus `provider_error` metadata when the provider is inaccessible.
- Sanitizes exposed provider error URLs by removing credentials, query strings, and fragments.
- Updates the artist-detail frontend to display API failures and owned-artist provider warnings.
- Replaces quote/style-sensitive source assertions with behavioural AST-based integration checks.

## Tests

Coverage includes:

- connection, timeout, TLS, HTTP, and JSON failures;
- wrapped clients and request-session access;
- explicit not-found versus search failure;
- fallback after a valid empty response;
- stopping fallback after a provider-access failure;
- successful recovery after an internal same-provider failure;
- an internal failure explaining an otherwise empty result;
- sanitization of credentials and query parameters in provider errors;
- Spotify no-auth ID and name-resolution behaviour;
- source-only hard failures;
- owned-library fallback with provider-error metadata;
- strict routing limited to the two web endpoints;
- preservation of the lenient legacy metadata facade;
- frontend propagation of provider errors and warnings.

Validation completed before the review-fix commit was pushed:

- Python syntax compilation;
- JavaScript syntax check;
- `git diff --check`;
- Ruff on all changed Python files;
- focused metadata, source-override, artist-source-detail, and web-integration tests.

## CI note

The branch was originally based on a `dev` commit containing Ruff B905 in `api/video/discover.py`. That unrelated issue has since been fixed in current `dev` with `zip(..., strict=True)`, so the PR merge result includes the correction from the target branch without adding video changes to this PR.


Nezreka:
hey, really solid work here. the empty vs error split is a real gap and you nailed the diagnosis. applied it locally, all 733 metadata tests pass. three things before i can merge — none of them structural, the design is right. they all boil down to the same rule: surface provider failures where a user is looking, change nothing anywhere else:

1. **recovered failures still error.** `call_discography_provider` raises if the capture saw *any* failure, even when the client recovered and returned real results. a client that hits a dead mirror then succeeds on the next one gets its catalog thrown away and a 500 instead (repro'd this locally — hifi does exactly this retry pattern). if the callback returns results, return them — a recorded failure should only explain emptiness, never override success.

2. **owned artists lose their library on a provider hiccup.** in `get_artist_detail` your error return sits before the `merged_discography = owned_releases` fallback, so an owned artist's page now 502s where it used to still show what they own. hard-fail is right for source-only artists, but for owned ones surface the error on a 200 alongside their library instead.

3. **scope leaks into background jobs.** `metadata_service` re-exports the strict versions and `repair_jobs/discography_backfill.py` imports from there, so a repair job silently goes strict too — while `core/metadata/__init__.py` still exports the lenient ones, so the two facades now disagree. i'd route just the two web endpoints through `discography_strict` explicitly and leave the facades alone.

smaller stuff, take or leave: exception strings can carry full request urls (some providers put keys in query strings) so sanitize before sending to the frontend, and the web-integration tests assert exact source strings from web_server.py down to quote style — those will break on any harmless refactor.

happy to merge once 1-3 are sorted. thanks for the thorough tests and writeup, this is a great direction. longer term i'd love the clients themselves to distinguish "no results" from "couldn't reach" so the proxy layer can shrink to a simple try/except — if you're interested that'd make a great follow-up.

ragnarlotus:
Thanks for the detailed review and for running the full metadata suite.

I agree with the three required changes, and I want to clarify the intended semantics before I update the branch because this PR is meant to be an intermediate step toward stricter source-authoritative lookups.

For point 1, I agree that a recovered internal attempt within the same provider must not be turned into an error. The rule I’ll implement is:

- callback returns releases -> success, even if an earlier mirror/request inside that same provider failed;
- callback returns empty and a provider-access failure was captured -> error;
- callback returns empty and no technical failure was captured -> valid empty result.

The important boundary is between retries/mirrors inside one provider and silently switching to a different provider. The longer-term goal is that, when the user explicitly selects a metadata source, that source becomes authoritative: internal retries for that provider may recover, but a failure of the selected provider must not be replaced with results from another provider.

I also agree with points 2 and 3:

- source-only artists should hard-fail on provider access errors, while owned artists should still return HTTP 200 with their owned releases plus provider-error metadata for the UI;
- the strict path should be imported explicitly only by the two web endpoints, leaving `metadata_service`, background repair jobs, search, import, and enrichment on the existing lenient behavior.

I’ll also take the two smaller suggestions: sanitize provider errors before exposing them to the frontend, and replace the quote/style-sensitive web-server assertions with less brittle integration checks.

Your longer-term suggestion to move the distinction into the provider clients themselves matches the intended follow-up exactly. This PR is the compatibility layer needed to preserve `empty` versus `error` at the user-facing boundary first; once clients expose that contract directly, the instrumentation/proxy layer can be reduced substantially.

Nezreka:
Beautiful! Thank you for doing this. I'm heading out for a couple days but will check on this first thing. Thank you again!

ragnarlotus:
Implemented the requested review changes and updated the PR description.

Changes now on the branch:

1. **Recovered same-provider failures no longer override success**
   - if an internal mirror/request fails but the provider later returns releases, those releases are returned;
   - a captured provider failure only becomes an error when the final result is empty.

2. **Owned artists keep their library on provider failure**
   - source-only artists still return the provider error;
   - owned artists now return HTTP 200 with their owned releases plus `provider_error` metadata for the UI.

3. **Strict scope is limited to the two web endpoints**
   - `core.metadata_service` is back to the existing lenient discography exports;
   - artist detail and download discography import `discography_strict` explicitly;
   - background repair and other metadata flows remain unchanged.

Also included the two optional improvements:

- provider error URLs are sanitized before reaching the frontend, removing userinfo, query strings, fragments, and possible API keys;
- the web integration checks now assert behaviour/import boundaries using the AST rather than exact source formatting.

Added regression coverage for recovered success, failure-tainted emptiness, URL sanitization, owned-library fallback, strict route scope, the lenient legacy facade, and frontend warning propagation.

The review-fix commit was created only after Python compilation, JS syntax checking, `git diff --check`, Ruff, and the focused metadata/web test set completed successfully.

Head: `0c067b8d4b5bfe3c89581be5394ce41a7cac988c`

ragnarlotus:
Thank you — and likewise, I really appreciate the amount of work, care, and time you put into SoulSync. Maintaining a project of this size while still taking the time to review contributions so thoroughly is a considerable effort. Your feedback was thoughtful and technically useful, and I’m glad I could contribute something meaningful to the project. I’ll build the follow-up from the merged version and keep it as a separate, focused PR.

PR #1055, opened 2026-07-20
Misc fixes from Library V2 work: indexer URL leak, retry-loop stalls, tag-writer false positives, track-number collisions

### LINKED ISSUE #1: Docker Support
I know this was already requested on Reddit but you can provide docker support?
Would love to try this app :-)

### LINKED ISSUE #3: No Qt platform plugin could be initialized on Debian 12
After installing python3, pip and the requirements via pip, I do get the following error when trying to launch the application:

```
vale@arr-apps:/docker/SoulSync$ /docker/SoulSync/python/bin/python3 main.py 
2025-08-13 08:16:12 - newmusic - INFO - Logging initialized with level: INFO
2025-08-13 08:16:12 - newmusic.main - INFO - Starting Soulsync application
qt.qpa.plugin: From 6.5.0, xcb-cursor0 or libxcb-cursor0 is needed to load the Qt xcb platform plugin.
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: wayland, wayland-egl, minimalegl, xcb, vnc, linuxfb, eglfs, vkkhrdisplay, minimal, offscreen.

Aborted
```
Getting this error both on a ssh session and on a terminal from the OS GUI.

Host OS is a ProxMox VM running Debian 12 with a full GUI.

nick2000713:
These are miscellaneous fixes that came out of building and stress-testing a larger library-management feature I'm working on against real libraries .. downloads, tag writing, imports, quality profiles. Every item below is a bug I actually hit (or found while reviewing the fix for one I'd hit), and I reproduced each one before touching the code. Grouped by area, each with what was broken and what's different now.

 I'm opening this separately, ahead of libary v2, these fixes stand on their own... so there's no reason to sit on them until then.

### Security — indexer URLs never reach the browser (P0-03)

**Before:** Search results for torrent/usenet candidates embedded the raw indexer download URL (magnet link or NZB link) directly in the `filename` field sent to the frontend. Some private-tracker/Prowlarr URLs carry API keys or signed, time-limited parameters — those were sitting in plain sight in the browser's network tab and local state.

**Now:** Search results carry an opaque, server-generated token instead. The download endpoint resolves the token back to the real URL server-side, in a TTL-capped in-memory store. A raw URL or a forged/expired token is rejected outright — the client can no longer make SoulSync forward an arbitrary URL to the download client.

**Follow-up fix:** the token store's `put()` captured its expiry timestamp *before* acquiring its lock, so under concurrent search traffic near the store's size cap, a `put()` call could evict the very token it just created and hand back a token that was already dead. Moved the timestamp capture inside the lock so insertion order and eviction order can't diverge.

### Downloads — stuck/failed jobs that actually succeeded

**Before (P2-21):** When a usenet/torrent client reports a completed job but never exposes the final `save_path` (SAB does this routinely — `Completed` flips in History before `storage` is populated), SoulSync fell back to the client's in-progress directory *the moment* a wait window elapsed — without checking whether the client's own post-processing (unpack, repair, move) was still writing into that directory. That could mean staging a partially-written directory.

**Now:** the fallback requires the directory's on-disk fingerprint (size/file-count/mtime) to be identical across two consecutive polls before it's trusted. If it's still changing, SoulSync keeps waiting, bounded by the overall timeout instead of the short window.

**Follow-up fix #1:** that stability check ran on the *client-reported* path — for split-container setups (SAB in a different container/host than SoulSync, using `usenet_path_mappings`), that path was never locally readable in the first place, so the fingerprint always read "unreadable" and could never stabilize. The download would then hang for the full outer timeout (6h) instead of recovering within ~2 minutes. Fixed by resolving the remote-mount path *before* fingerprinting, not just on the final return value.

**Follow-up fix #2:** an in-progress directory that exists but is currently empty (created by the client, not populated yet) fingerprints as `(0 bytes, 0 files)` on every poll — two empty reads in a row looked identical to "stopped changing." Added a floor requiring at least one file before a fingerprint counts as stable.

**Also:** torrent/usenet downloads used to fall into the same retry-budget bucket as Soulseek peers, so an exhausted torrent search could burn through budget that should've been separate from a parallel Soulseek attempt. They now get their own budgets — and a stale, separate copy of the source-name list in `web_server.py` that still bucketed them under "soulseek" for UI display got synced to match.

### Async bridge — Python 3.14 compatibility without losing concurrency

**Before:** `run_async()`, the single shared bridge every sync call site uses to run a coroutine (download status polls, search, connection tests, web request handlers — ~150+ call sites), stopped reliably starting its background event-loop thread on Python 3.14.

**The original fix's side effect:** it replaced the shared `run_forever()` event loop (which lets many coroutines from different threads run concurrently, each yielding at its own `await` points) with a single-worker queue that runs one coroutine to *full completion* before starting the next — silently serializing every `run_async()` caller in the whole process. Confirmed by timing it directly: 3 threads each doing a 1-second async sleep took ~1 second concurrently before, ~3 seconds serialized after. In practice that means one slow network call (an unresponsive indexer, a stalled client) would head-of-line-block every unrelated download's status polling, search requests, and connection tests until it finished.

**Now:** back to the concurrent `run_forever()` model, keeping only the actual startup-race fix — a readiness signal (`loop.call_soon(ready.set)`) that only fires once the loop is genuinely pumping, so the first caller can't submit before the loop is ready without giving up concurrency to get there.

### Import pipeline — track numbers

**Before (§16.3(a)):** when no source (metadata, filename, embedded tag) carried a track number, the pipeline defaulted every un-numbered track in an album to track 1 — so a whole album with missing numbers collapsed onto one slot, and the library showed one track present and the rest "missing."

**The fix:** as a last resort, use the file's sorted position among its audio siblings on disk — the reasoning being that album-bundle downloads stage all of an album's files into one directory together, so sorted order is at least a distinct, if imperfect, fallback.

**Follow-up fix:** that reasoning only holds for genuine album-bundle downloads. A plain Soulseek download instead lands in a *shared* flat directory that can hold other, unrelated in-flight downloads at the same time — so an un-numbered single track could get assigned a position among files from a completely different download. Restricted the fallback to `is_album_download` contexts, which also incidentally closes a race where concurrent import workers processing the same folder could see a shrinking sibling list (as files get moved out mid-processing) and assign colliding numbers.

### Tag writer — false-positive retag warnings, and closing the gaps in that fix

**Before:** the "Re-tag" preview flagged files as needing a rewrite purely because of formatting differences — a full ISO timestamp on the file vs. a plain date in the DB, or a rich multi-genre tag getting flagged for "changing" to a generic single genre pulled from an enrichment pass.

**The fix:** normalize dates before comparing, and don't flag a genre change when the DB's value is already contained in the file's existing genre list.

**Follow-up fix #1:** the date-normalizer's regex required a time-of-day component to match at all, so the single most common shape of this exact problem — a bare `YYYY-MM-DD` on one side vs. `YYYY-MM-DDT00:00:00` on the other — was never actually recognized as equal, leaving the false positive it was written for unfixed for that case.

**Follow-up fix #2:** the "keep the existing date" logic used string *length* as a proxy for "more specific," so a stale, genuinely wrong existing date (`2023-01-01`, same year, wrong month/day) could permanently block a real, shorter correction (`2023-09-05`) from ever being written.

**Follow-up fix #3:** the genre-diff suppression only existed in the *preview* — `write_tags_to_file` had no matching check. A batch write (which pre-filters files using the preview's verdict) would silently skip a file whose only real change was a genre narrowing, while a single-track write (which calls the writer directly, bypassing the preview) would still apply it — two different outcomes for the exact same data, depending only on which button you clicked. Extracted the subset check into a shared helper so the preview and the actual write always agree.

### Also in this batch
- SABnzbd category whitespace: the new connection-test category check strips the configured category before validating it against SAB's category list, but the adapter never stripped it before submitting — a category with incidental whitespace could pass the check and still get silently rewritten to `*` by SAB on the real download.
- `delete_quality_profile` now guarantees a default profile survives any deletion, even if `is_default` bookkeeping was ever left inconsistent by an older bug or a hand-edited DB.
- Simple/direct downloads (no provider enhancement) now get title/artist/album written into their tags instead of staying untagged.
- Automation progress is clamped to 0–100%, so a counter race under load can't show >100% in the UI.

### DIFF
--- a/core/automation/progress.py
+++ b/core/automation/progress.py
@@ -30,6 +30,20 @@ progress_states: dict[int, dict] = {}
 progress_lock = threading.Lock()
 
 
+def _clamp_percent(value: Any) -> Any:
+    """Clamp a reported progress value to 0-100.
+
+    Callers compute `progress` from counters (processed/skipped/failed vs. a
+    precomputed grand total) that can drift past the nominal total under
+    races or double-counting, so this is the one place that guarantees the
+    polled/pushed state never shows a bar or label outside 0-100.
+    """
+    try:
+        return max(0, min(100, round(float(value))))
+    except (TypeError, ValueError):
+        return value
+
+
 def init_progress(automation_id: int, automation_name: str, action_type: str) -> None:
     """Initialize progress state when an automation starts running."""
     with progress_lock:
@@ -70,6 +84,8 @@ def update_progress(
                 state['log'].append({'type': kwargs.get('log_type', 'info'), 'text': v})
                 if len(state['log']) > 50:
                     state['log'] = state['log'][-50:]
+            elif k == 'progress':
+                state[k] = _clamp_percent(v)
             elif k != 'log_type':
                 state[k] = v
         if kwargs.get('status') in ('finished', 'error'):
--- a/core/connection_test.py
+++ b/core/connection_test.py
@@ -342,9 +342,19 @@ def run_service_test(service, test_config):
                     if client_type == "sabnzbd":
                         return False, "SABnzbd needs both URL and API key."
                     return False, "NZBGet needs URL, username, and password."
-                if run_async(adapter.check_connection()):
-                    return True, f"Connected to {client_type}"
-                return False, f"{client_type} probe failed — check URL, credentials, and that the client is running."
+                if not run_async(adapter.check_connection()):
+                    return False, f"{client_type} probe failed — check URL, credentials, and that the client is running."
+                if client_type == "sabnzbd":
+                    category = str(config_manager.get(
+                        'usenet_client.category', 'soulsync') or 'soulsync'
+                    ).strip() or 'soulsync'
+                    if not run_async(adapter.category_exists(category)):
+                        return False, (
+                            f"Connected to SABnzbd, but category '{category}' "
+                            "is not configured there. Add the same category "
+                            "in SABnzbd before enabling acquisition."
+                        )
+                return True, f"Connected to {client_type}"
             except Exception as e:
                 return False, f"Usenet client connection error: {str(e)}"
         elif service == "torrent_client":
--- a/core/download_plugins/album_bundle.py
+++ b/core/download_plugins/album_bundle.py
@@ -401,6 +401,47 @@ def snapshot_incomplete_path(path: str) -> Optional[tuple]:
         return None
 
 
+def incomplete_path_stability_check(
+    raw_path: str,
+    prior_path: Optional[str],
+    prior_snapshot: Optional[tuple],
+    *,
+    snapshot_path: Callable[[str], Optional[tuple]] = snapshot_incomplete_path,
+    resolve_path: Callable[[str], str] = lambda p: p,
+) -> tuple:
+    """P2-21 stability gate for the ``incomplete_path`` fallback, shared by
+    ``poll_album_download`` and the usenet single-track poll loop.
+
+    ``raw_path`` is the client-reported in-progress directory, which may
+    live inside the client's OWN container/mount and not be directly
+    readable here. ``resolve_path`` (typically ``resolve_reported_save_path``)
+    is applied BEFORE snapshotting so the stability check — and the path
+    returned on success — both operate on somewhere this process can
+    actually read; snapshotting the raw path meant a remote-mounted
+    directory always read as unreadable (``None``) and could never
+    stabilize, stalling the fallback until the outer poll deadline instead
+    of the intended ~120s window.
+
+    A snapshot with zero files never counts as stable either: the client
+    may have just created an empty in-progress directory a moment before
+    starting to populate it, and two empty reads in a row would otherwise
+    look identical to "writes have stopped".
+
+    Returns ``(stable, resolved_path, new_snapshot)``. Callers should
+    remember ``resolved_path``/``new_snapshot`` as ``prior_path``/
+    ``prior_snapshot`` for the next poll regardless of ``stable``.
+    """
+    resolved = resolve_path(raw_path) or raw_path
+    current = snapshot_path(resolved)
+    stable = (
+        current is not None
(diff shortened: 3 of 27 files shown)


Nezreka:
hey, thanks for this one. the level of care here is honestly impressive, every fix came with the follow-up fixes for its own edge cases and real tests to back it all up. stuff like the indexer url tokens and the async bridge fix are things that quietly make the whole app better for everyone. ran the whole batch against current dev, everything green, and the fixes hold up exactly like you described them. also appreciated you making those blocklist tests hermetic while you were in there.

two small things and it's in:

1. the simple-download tagging overwrites tags instead of just filling them in. your description says "instead of staying untagged" and that's the behavior i want, but write_tags_to_file will stamp over a real existing title/artist/album too, and soulseek search titles are parsed from filenames so they can be messy. only write a field when the file's value is empty or a placeholder.

2. the stability gate: if the in-progress path can never be read at all (no path mapping set up, not just an unresolved one), the old code failed fast at staging. now it quietly retries for up to 6 hours before saying anything. maybe cap it, after a handful of unreadable polls in a row just give up with the old error? curious what you think.

fix 1, weigh in on 2, and i'll happily merge. thanks again for not sitting on these until v2.

nick2000713:
Fixed both:

  1. Tagging now reads the file's current tags first and only fills title/artist/album when the existing value is blank or a placeholder — a real tag is never overwritten, regardless of how "trustworthy" the search-result value looks.

  2. Added a bounded counter to the stability gate: if `incomplete_path` can't be read at all for a handful of consecutive polls (not just "still changing"), it now gives up with the same error the pre-stability-gate code used, instead
  of silently waiting out the full 6h timeout.

ready to merge..

PR #1062, opened 2026-07-22
Library overhaul

nick2000713:
# Library Manager v2: a real, Lidarr-equivalent library manager on SoulSync's own pipeline

SoulSync's library page was always a **read-only mirror** of whatever Plex/Jellyfin/Navidrome reported back. it could only show you what a media server had already decided, and nothing else: no "monitor this artist", no missing tracks, no way to pick a release when auto-grab got it wrong, no repair job you could point at one artist.

Library v2 replaces it with SoulSync's own library. not a prettier mirror.. **a catalogue you can rebuild from**, with the front end and the decision layer Lidarr has, sitting on the search, download, quality and tagging pipeline SoulSync already had.

> **this is not opt-in anymore.** it started behind a `features.library_v2` flag in `config/config.json`. that phase is over: lib2 **is** the library now. no flag, no json file, nothing to toggle. the old key is read once, logs that its deprecated, and is ignored.

---

## What changes

| | old library | Library v2 |
|---|---|---|
| what it is | a mirror of the media server | SoulSync's own catalogue |
| needs a media server | yes, for everything | no, not even for artwork |
| monitoring | artist-level watchlist | artist, album **or** single track |
| tracks you dont own | invisible, just a gap | real rows with real titles, monitorable |
| picking a release | whatever auto-grab chose | interactive search, you pick |
| quality check | after the download | **before** you grab |
| song by two artists | showed under one | stored once, shows under both |
| FLAC + MP3 of one song | one was invisible | many files per track, one primary |
| deluxe vs original | whichever imported first won | album idea and pressing are separate |
| download status | "N downloading" for everything | queued / searching / downloading / processing |
| history | download rows, fuzzy name match | one feed: grabs, checks, retries, quarantine, imports, moves, deletes |
| your manual tag edits | a refresh could overwrite them | marked as yours, refresh skips them |
| delete | database rows only | real files, with a preview first |
| repair tools | fixed legacy rows the library never saw | all 25 write back to the library |

---

## How it's built

its **not** a second download/search/tagging system. its a new model and a new front end talking to the pipeline that was already there:

```mermaid
flowchart TB
    subgraph UI["Library v2 UI"]
        A[Artist / Album / Track view]
        B[Interactive Search & Manage Tracks]
        C[Quality Profiles & Re-Tag Preview]
    end

    subgraph Bridge["Monitoring mirrors into the existing systems"]
        D[Artist monitor → Watchlist]
        E[Album/Track monitor → Wishlist]
    end

    subgraph Core["SoulSync's existing pipeline: unchanged, reused, not duplicated"]
        F[Multi-source Search]
        G[Download Orchestrator]
        H[Quality Gate + AcoustID]
        I[Tagging / Post-Processing]
    end

    UI --> Bridge --> Core
    Core -. same auto-download machinery keeps running .-> Bridge
```

"monitor this artist" doesnt invent a new queue, it adds the artist to the existing Watchlist, the same one the old page used. same for Wishlist.

there are three ways a song gets in and only one of them is a download. manual grab is you picking a result, automatic search picks for you, and if it finds nothing or the download dies the track goes on the wishlist which retries later. import is different because nothing told it what the files are, so it identifies them first and parks anything it isnt confident about instead of guessing:

```mermaid
flowchart LR
  MG(["manual grab"]) --> DL["download"]
  AS(["automatic search"]) --> DL
  DL -.->|"failed or\nnothing found"| WL[("wishlist")] -.->|"retries later"| DL
  IM(["import: you drop\nfiles in staging"]) --> ID["identify by tags,\nfilename, AcoustID"]
  ID -.->|"not confident\nenough"| RV[("parked for you\nto confirm")] --> CHK
  ID --> CHK
  DL --> CHK
  CHK["checks: integrity,\nsilence, AcoustID"] -.->|"fails one"| Q["quarantine"]
  CHK --> TAG["write tags"] --> P["build path"] --> MV["move into library"] --> CAT[("catalogue")]
```

from the checks onward its the exact same code for all three, and the catalogue row is what proves it landed. a file sitting on disk without that row does not count as imported.. thats an explicit guard now, not something we hope happened.

after that, four things can touch that row, and each only does its own job instead of all of it happening inside reorganize:

| step | reads | writes |
|---|---|---|
| you, editing a field | nothing, its your call | that one field, and marks it as yours |
| manual match | your pick of the correct release | just the provider id on the row |
| refresh | that id, then the provider | title, track numbers etc, skips any field marked yours |
| re-tag | the row | the file's tags, same skip rule |
| reorganize | the row | the file's path, moves it |

---

## Where it lands vs. Lidarr

| Capability | Old Library page | Library v2 | Lidarr |
|---|:---:|:---:|:---:|
| Read-only mirror of the media server | ✅ (that's *all* it is) | ❌ (own library) | ❌ |
| Works with zero media server attached | ❌ | ✅ | ✅ |
| Monitor artist / album / track individually | ❌ | ✅ | ✅ |
| Missing-tracks view with real titles, not gaps | ❌ | ✅ | ✅ |
| Wanted: missing + cutoff-unmet views | ❌ | ✅ | ✅ |
| Interactive search → pick a specific release | ❌ | ✅ | ✅ |
| Quality verdict shown before you grab | ❌ | ✅ | partial |
| Live download state per row and per album | ❌ | ✅ | ✅ |
| One history feed per artist/album/track | ❌ | ✅ | ✅ |
| Album idea vs. concrete pressing | ❌ | ✅ | ✅ |
| Multiple files per track (FLAC + MP3) | ❌ | ✅ | ✅ |
| Quality profiles enforced at every stage | partial | ✅ | ✅ |
| Full discography + monitor future releases | ❌ | ✅ | ✅ |
| Re-tag preview (file tags vs. library) | ❌ | ✅ | ✅ |
| Per-field ownership of your manual edits | ❌ | ✅ | partial |
| Manage duplicate single/album versions | ❌ | ✅ | ✅ |
| Per-artist scoped repair jobs | ❌ | ✅ | ✅ |
| Delete removes the actual files, safely | ❌ | ✅ | ✅ |

a few Lidarr things i didnt take on purpose: inline table editing (modals are safer), an artist mass editor, metadata profiles, a calendar, a blocklist UI, A-Z source headers, and "Add Artist" as its own page. Search and the Watchlist are the entrance already, a second one would just confuse people. Top Tracks and batch discography download are out too.

---

## What's new

- **interactive search:** search every configured source for one artist, album or track and pick the result yourself. each row shows its source family (p2p / usenet / torrent / streaming), the numbers that decide whether a grab completes (Soulseek free slots and queue length, usenet indexer and grab count) and the release age as `3d` / `8mo` / `2.1y`. every row is measured against the quality profile **before** you grab, with the same rules the import gate applies after. and its source-aware on purpose, so a source that never reports bit depth doesnt look like it fails every hi-res profile.
- **live download state:** a row says what its really doing, queued, searching, downloading or processing, and an album rolls its tracks up into the same four. before, everything in flight just said "N downloading", so three tracks waiting behind a busy client claimed to be downloading.
- **one history per entity:** grabs, checks, retries, quarantine, import outcomes, canonical relinks, file moves and deletes in a single feed. it used to be download rows only, matched by a fuzzy artist-name comparison.
- **your edits win:** an edited field is stored *and* marked as yours. refresh skips it, re-tag shows both values and lets you pick per field, and a bulk apply counts them first: "23 findings, 4 of them yours".
- **quality profiles that mean one thing:** one table, no shadow copy. search ranking, the AcoustID skip, the import gate and the upgrade scanner all resolve the same profile live, so a change in Settings applies everywhere at once.
- **discography and discovery:** browse an artist's full provider catalogue, bulk monitor, let new releases appear on their own. an artist you own nothing from opens inside Library v2 and stays read-only while you browse, so it doesnt fill the library with artists you never asked for.
- **manage track files:** see single-vs-album duplicates, unlink them, or move a file onto the correct version without touching disk.
- **safe delete:** a preview that changes nothing first, then a journalled operation. only paths inside your configured `library.music_paths` roots can be deleted at all, unknown mounts and symlink escapes fail closed.
- **repair tools that actually land:** all 25 migrated onto library subjects and scoped to one artist where the job allows it. otherwise a tool could fix the file while the library kept showing the old cover, tags or missing state forever.
- **artwork without a media server:** embedded cover first, then a provider, cached locally. a pure SoulSync install looks exactly like one with Plex attached.

---

## How to try it

back up your data dir with the container stopped, start it, let the migration run. thats it, theres nothing to enable.

the migration runs in the background so the UI comes up first, logs its stage and counter every 30 seconds, and resumes from a checkpoint if you restart in the middle.

---

I screwed up — a force-push (git history cleanup) automatically closed the old PR #1025, and GitHub won't let it be reopened since then ("branch was force-pushed or recreated"). The branch/code is unchanged, only the PR container itself is gone. Here are the comments from last time so nothing gets lost:

**@Nezreka** (2026-07-14):

> I'm excited to test this! Would this replace the entire library page or one of the views that exist in it. Standard view was nothing special but enhanced view was more library focused for manipulation

**@nick2000713** (2026-07-14):

> It’s still a long way from being finished, so there’s a lot missing and a lot left to implement.
> 
> The idea is that it will eventually replace both the standard library and Library Enhanced, although it’s mainly meant as a replacement for Library Enhanced for power users.
> 
> If you notice something missing, definitely let me know. There’s still plenty I want to add.
> 
> Personal note: I actually started working on this before SoulSync expanded into movies as well. Personally, I’d rather keep focusing on music for now. I feel like the movie/TV (and even YouTube) space is already covered really well by projects like Sonarr, Radarr and TubeSync, which are, in my opinion, still superior in those areas. Because of that, I haven’t really designed or optimized Library v2 with those use cases in mind.

**@Nezreka** (2026-07-14):

> Sounds great! I'm just messing with it now, importing my library. Excited to see how it all works. Yeah you focus on anything you want to, the entire video side is very fresh with more work to be done. Music and video side are fully disconnected though so no worries you will impact anything. I've been daily driving it replacing radarr, sonarr, kometa and ytdl-sub. takes a day or so to enrich the library for the functions to work though. 

**@Nezreka** (2026-07-14):

> I'm noticing that the import is taking quite a bit without any live status. Been about 30 min without much happening but it says it is importing. I have a 320,000 track library though.

**@nick2000713** (2026-07-14):

> Oh Wow, definitely never tested it at 320k tracks 😅 Good news, it's not actually stuck.. the DB import itself is fast, but right after that it silently kicks off an enrichment pass that fetches tracklists + artwork per album from Spotify/Deezer, one by one. At your scale that's tens of thousands of sequential API calls, so itjust grinds for a while.
> 
> The actual bug is that none of that shows up in the UI — it just sits on "Importing…" forever with zero 
> indication anything's happening, even though the backend is tracking real progress under the hood. Logged it as P2-25, will get real progress reporting wired up. For now if you tail the backend logs you should see it steadily crunching through albums.
> 

**@nick2000713** (2026-07-14):

> To be completely honest, the import still needs quite a bit of work. Several features from the old library haven’t been implemented yet, and not all information that was available in the old library is displayed in Library V2 yet.
> 
> If you’d like to try it, feel free to do so, but keep in mind that the import is still a work in progress.
> 
> For example, manual artist and track matching, metadata enrichment, and the detailed match overview from the old library are still on the roadmap. Some source information is not shown in Library V2 yet, but I plan to bring all of these features back.
> 
> I’m aware that the new library hasn’t reached feature parity with the old one yet, but that’s the goal. I’ll also be revisiting the import process to make it more complete and efficient.

**@Nezreka** (2026-07-14):

> > I’m aware that the new library hasn’t reached feature parity with the old one yet, but that’s the goal. I’ll also be revisiting the import process to make it more complete and efficient.
> 
> You're killing it, man. I appreciate your work. Take your time, let me know if you ever want me to pop in and check things. Moving SoulSync away from the reliance of media servers is necessary for other things planned down the line anyway.

**@nick2000713** (2026-07-16):

> Hey @Nezreka, I've worked on optimizing the import pipeline. It is now parallelized (for artwork, tracklists, and tag pre-caching) and runs significantly faster. Several bugs (like track-number healing and missing ReplayGain/Lyrics features after legacy import) have also been resolved.  Feel free to try it out and let me know what you think.. 
> 
> 

**@nick2000713** (2026-07-18):

> Hey @Nezreka, 
> 
> I've worked again on the import pipeline and resolved many many bugs. 
> 
> Several bugs are still known... I still have problems when the artist hasn't the right metadata or has duplicate artists, but I think I have it now somewhat under control. 
> 
> I also fixed the issue where newly split component artists were automatically set to monitored/watched instead of just the respective track. Now they correctly inherit the monitored status of the parent combined artist.
> 
> I would recommend to reset everything and start the import again since many many bugs are now resolved. It's better if you reset and start fresh, maybe now on a smaller library ;) 
> 
> If you want, you can execute this command in the browser. This is how you can reset it, but be aware it also removes everything on your wishlist/watchlist:
> 
> ```javascript
> (async () => {
>   console.log('Hole Liste aller Künstler...');
>   let page = 1;
>   let allArtistIds = [];
>   
>   while (true) {
>     const res = await fetch(`/api/library/v2/artists?page=${page}&limit=500`).then(r => r.json());
>     if (!res.success || !res.artists || res.artists.length === 0) break;
>     allArtistIds.push(...res.artists.map(a => a.id));
>     page++;
>   }
>    
>   console.log(`${allArtistIds.length} Künstler gefunden. Löschvorgang gestartet...`);
>   
>   for (let i = 0; i < allArtistIds.length; i++) {
>     const id = allArtistIds[i];
>     await fetch(`/api/library/v2/artists/${id}`, { method: 'DELETE' });
>     console.log(`[${i + 1}/${allArtistIds.length}] Künstler mit ID ${id} gelöscht`);
>   }
>    
>   console.log('Bibliothek vollständig geleert! Lade Seite neu...');
>   location.reload();
> })();
> ```
> 
> After the import and caching is done, I would recommend to manually start the "Reconcile unmapped artists" job. It resolves collaboration names into the real component artists and several other things. 
> 
> Try it out and tell me what you think, thank you!

**@nick2000713** (2026-07-18):

>  I am now working on the diverse tools that work with the old library, etc. together. For me, every feature I want and the old library had is implemented now. It's easier for me to build solely on the new library (Library v2), so I have to change/modify nearly every tool because it depends on the old library and database tables. For me to develop, it's easier to solely concentrate on Library v2 and migrate every tool over. Therefore, with the new code, the old library will be removed so it makes it easier going forward. If something is missing for you, tell me now...
> 
> And another thing is, I have found several bugs anyway and solved them in the download pipeline. Like when you approve a track in the quarantine, it would later be detected as an orphan file because the metadata after a given time weren't present so it wasn't written in the database. Such bugs and several others are also resolved because I stumbled upon them. @Nezreka 

**@Nezreka** (2026-07-19):

> Okay, I'm starting to understand this better now that I've got it running on a test library. My big question: what happens to the artist-detail pages?
> 
> Right now those are the only thing that lets me pull up an artist's full discography for artists I *don't* own yet — as far as I can tell Library v2 only works on artists already in my library, so I want to understand how that "find an artist I have nothing by and start monitoring them" flow survives once the old library is gone.
> 
> I'll be honest, this is getting a little big lol  — scope mostly. But I'm willing to see it through and test as much as you need but also it's early and I haven't seen the full picture yet. I just want us on the same page about what's getting **replaced** (a page goes away) vs. **rewritten** (same page, new plumbing underneath) before it gets expensive to change. 
> 

**@nick2000713** (2026-07-19):

> Thanks for testing it so thoroughly. And don't worry, we absolutely don't have to rush this. I completely agree that before something this big gets merged, the architecture has to be right. I'd rather take more time now than end up with something we regret later.
> 
> The reason this became much larger than I originally expected is that I started Library V2 as an opt-in feature. Because of that, I introduced separate database tables instead of replacing the existing ones. That worked fine at first, since Library V2 was just a new view sitting on top of what already existed. But once I actually started testing it against a real library, I kept running into background jobs that quietly assumed a track could only ever live in the old tables. As one example: the lyrics-fetch job resolved a track's file path through the old library tables, and once tracks could also exist purely in Library V2's own tables, that same resolver silently returned wrong paths for V2-only tracks. I fixed that, and a few days later hit the exact same class of bug independently in the ReplayGain job, because it had its own separate copy of similar path-resolution logic. And it wasn't just those two — there are 33 separate repair/maintenance tools (quality scanning, tagging, cover art, orphan detection, and so on) that all made the same old-tables assumption in one place or another. At that point I basically had two choices:
> 
>   - keep both libraries alive and write synchronization between the old and new tables for every one of those tools, or
>   - migrate the tools to Library V2 directly.
> 
> I chose the second option because synchronization is much more error-prone than having a single source of truth, and because fixing the same class of bug over and over, one tool at a time, wasn't sustainable. In hindsight, that's also why this PR grew much more than I originally anticipated.
> 
> I understand your concern about the artist pages, but I don't think anything you're actually worried about is being removed — this is a wording problem, not a scope problem. To be clear about what's actually going away: it's only the old Library page itself. Search, Discover, the artist-detail page, Watchlist, and Wishlist are separate pages and none of them are being touched by this migration.
> 
> Given that, I have to admit I'm not fully sure I still see the problem — you can already search for an artist you don't own today, open their page, browse their discography, and add them to the Watchlist or download straight from there, all without that artist ever being in the library. That flow doesn't sit on the page I'm removing, so it keeps working exactly as it does now. If there's a specific part of it you're worried about beyond the wording, let me know and I'll take a closer look.
> 
> The old Library page itself is essentially replaced by Library V2, so I don't see much value in keeping two versions of the same thing around. Everything else stays as is for now.
> 
> Also, while working on this I stumbled across a surprising number of unrelated bugs in the existing pipeline. Some of those fixes probably don't even belong in this PR and could be split into separate pull requests to make this one easier to review.
> 
> On top of that, the playlist view — we can strike that entirely from scope. I'm not attached to it and it's not something I need in this PR.


nick2000713:
@Nezreka 
its been a long time since the last update and i improved the library alot since..

First; i initially misunderstood your concern about how discovering new artists would work if the legacy library would be gone.. i was mainly looking at artists that existed inside the new catalogue, but the real missing case was an artist that you find through Search but dont own anything from yet..

This is now fully implemented inside Library V2. If you come from Search you stay inside the new library and get a similar view to the old one, with the full discography, the legacy cards, the filters, the richer artist header, listeners, plays and Top Tracks.. there is also a switch between the normal Library V2 table and the old card view.

The discovery view stays read-only while you are only browsing, so it does not fill the database with artists that you dont actually want. Only a real action like opening a release or bookmarking something creates the needed Library V2 entries, and bookmarking one Top Track monitors only this track and not the complete album or artist..

I also researched why the covers in the old view loaded so much faster. The old library sent the provider image directly to the browser, while Library V2 first tried to build every cover locally through a limited worker pool. Now releases that are only part of the discovery view use the fast provider image directly, while owned and monitored releases still keep the local cache as the real long-term source.. so the complete old artist/discovery flow is now replaced and there is no Search path that still needs to send a user into it.

Second: I totaly agree this got realy realy big and there are still always things that could be improved.. but now i made multiple bug runs, test runs and also checks against a copy of my real database and real music files, so im very very confident about this work and im fully committed to make this much better in the future too..

## Why this branch got so big

I also want to explain the size because i agree that looking only at the complete diff is alot.. measured from the shared `dev` merge-base to the current branch, it changes **483 files** with around **130,690 added lines** and **16,270 removed lines** over 96 commits.

These are Git diff additions, so a line that was moved or rewritten also counts as a new line.. but the breakdown gives a much better picture of where the size actually comes from:

| Area | Files | Added lines | Removed lines | Why it is there |
|---|---:|---:|---:|---|
| Library V2 backend and API | 65 | **30,431** | 0 | The new catalogue, schema, queries, artist/release/track/file models, artwork, monitoring, Wanted, history, import, editions, metadata and all Library V2 endpoints |
| Persistent acquisition pipeline | 30 | **11,614** | 0 | Requests, candidates, grabs, retries, client adoption, bundle matching, quarantine/import correlation and restart-safe history |
| Library V2 frontend | 15 | **19,895** | 30 | The complete new React Library, artist/release/track views, search, actions, columns, modals, discovery and artwork handling |
| Tool and repair migration | 39 | **5,044** | **8,933** | Moving the old tools onto Library V2 subjects and lifecycle while removing retired or duplicated legacy repair logic |
| Shared pipeline integration | 75 | **6,013** | 1,087 | Connections into the existing import, search, Wishlist, metadata, download-client and database code |
| Tests | 246 | **49,457** | 6,196 | Backend, acquisition, repair, frontend and regression coverage for the new behaviour and dangerous failure cases |
| Documentation | 6 | **8,166** | 1 | Guide, feature decisions, status, issue/root-cause register and migration documentation |
| Build/config/other | 7 | **70** | 23 | Small supporting changes |

So the actual new Library V2 product implementation is around **61,940 added lines** when the backend/API, persistent acquisition layer and frontend are counted together.. almost **50,000 more lines are tests**, and another **8,166 lines are the documentation and audit trail**.

The backend is large because this is not only a new screen over the old database.. Library V2 adds its own normalized catalogue with stable artist, release, edition, track and file identities, multiple files per track, real file lifecycle state, Wanted/monitoring rules, provider snapshots, metadata overrides and history.

The acquisition part is separated in the table because it is another major reason for the size.. Search and Download itself are still reused, but Library V2 needs persistent requests, candidate/grab/import correlation, retries, quarantine state and restart-safe adoption of external clients so the UI can explain what happened after a restart and does not invent a second live queue.

The frontend is almost 20,000 lines because the new Library is a complete React/TanStack page instead of another patch inside the old vanilla-JS library.. it includes all artist, album, single, track, Wanted, file, history, search, quality, metadata, maintenance and discovery views plus the actions and dialogs around them.

The tools are not the main reason for the added size.. their migration added around **5,044 lines**, but it also removed around **8,933 lines**, so this area is around **3,889 lines smaller in the final diff**. The removed code is mainly retired or duplicated legacy repair behavior that would otherwise leave two different repair systems.

The tests are so large because nearly every new state transition has more than one side.. a file move for example is not only “did the file move”, it also has to prove that the Library V2 path, primary-file choice, Wanted state and history stay correct if the operation fails, repeats or sees more than one file. The same is true for monitoring, provider identities, imports, quarantine, artwork and restart adoption. This is why the branch contains almost as much test code as product code instead of only a few UI tests.

## A few things to clarify why the tools and jobs have been modified

Before this branch existed, none of the existing tools knew anything about Library V2.. they worked with the legacy library tables, the filesystem or both, because that was the only library that existed.

During this branch Library V2 became its own DB-centered catalogue and the new source of truth for artists, releases, tracks, files, monitoring and file state. Because of this, the existing tools had to be migrated step by step.. otherwise they could still find or fix something in the old library or directly on the file, but the new library would never know that anything changed.

This is also what the “23 of 25 jobs” result means.. it does **not** mean that 23 jobs supported Library V2 before this branch. It means that at the time of the final tool audit, the work inside this branch had migrated 23 of them and the audit found the last two jobs that were still reading only legacy data.

The main goal was also not to create a second set of tools.. i kept the existing tool behavior and connected it to the new catalogue. A native fix now needs to resolve the correct Library V2 subject, apply the existing repair, rescan only the affected files, update artwork or Wanted when needed and write the result into the history. Without this, a tool could say “fixed” while Library V2 continued to show the old cover, tags, path, verification or missing state forever..

### Cover Art Filler, Metadata Gap Filler and Library Re-tag

These tools originally only understood legacy album/track rows and the existing file paths. They were migrated to native Library V2 album, track and file subjects, but this exposed another problem: changing the real file did not automatically update the new catalogue snapshot.

For example, Cover Art Filler could correctly embed a cover into the audio file, but Library V2 could still show the same cover gap afterwards because its cached tag state was never refreshed. The normal Re-tag path also looked only at text changes, so a file with correct title/artist tags but a missing cover was skipped and the UI still said “Tags written”..

Now a missing cover is its own real write reason, the preview can show `None -> Available`, a cold cover can be loaded through the shared artwork resolver and the affected Library V2 file is rescanned afterwards. Ogg and Opus embedded pictures are also read through the same representation that the cover writer uses, so they no longer produce a permanent fake cover gap.

Metadata Gap Filler was migrated in the same way: native Library V2 subjects, correct provider-ID positions and a targeted rescan after the fix. Two native subject builders also had one missing legacy compatibility column, which shifted the remaining provider values and could crash the complete job on a real upgraded database.. this was only visible after running it against a migrated production DB shape and is now fixed.

Library Re-tag keeps the existing common tag writer instead of adding another V2 tag engine. The difference is that the preview and apply path now get their data from Library V2 and the result is written back into the file state that the new UI actually reads.

### Refresh & Scan

Refresh & Scan was migrated to use `lib2_track_files` as the file list instead of treating the legacy library as its catalogue. It checks the real file existence, audio quality, embedded tags, features and the `SOULSYNC_VERIFICATION` marker and then stores these observations in Library V2.

It intentionally does not become another provider metadata tool.. Enrich and Discography Refresh own that work. Mixing provider calls into a normal file scan would make every scan slower and would duplicate the metadata logic. So the change here was to migrate the file inspection and make its real scope clear, not to make it responsible for every type of refresh.

### Genre Cleanup and Comma Artist Splitter

These were the last two jobs found by the audit that were still really reading only the legacy artists and albums. During the earlier migration work in this branch their registered job definition was changed to say `lib2`, but their internal queries were missed and stayed on the old tables.

Both now have native Library V2 readers and identities, but i did not change what the jobs are supposed to do.. Genre Cleanup still only removes genres that are not allowed and does not invent new genres. Comma Artist Splitter still uses the same whitelist and provider checks before it splits anything.

The important difference is that they now see the complete Library V2 catalogue and that Comma Artist Splitter retags the affected Library V2 files afterwards. Genre Cleanup stays metadata-only, because rescanning a complete artist discography for a cleanup that only changes catalogue metadata would be unnecessary and very expensive.

The missing-genre situation itself is intentionally not hidden: none of the configured providers supplied reliable album genres for the real affected albums. Genre Cleanup is an **updater/cleaner**, not a genre procurement tool, and i did not add a weak artist-genre fallback only to make the gap disappear in the UI.

### Library Reorganize and Path Drift Reconcile

Library Reorganize had to be migrated because its old planner and queue knew legacy track IDs and legacy file paths. The existing move engine was kept, but the findings and the result now need an explicit Library V2 identity and the final move has to update `lib2_track_files.path`, the compatibility path and the history together.

One bug found during the audit came from the first migration of this job inside the branch: it read native rows but wrote their IDs as normal numbers. A normal number can also be a legacy ID, so the finding could attach an unrelated Library V2 track that happened to carry this number as its old back-reference. The job now writes an explicit `lib2:` identity and the exact Library V2 subject details.

It also no longer reports success when the physical rename worked but the catalogue update failed. In that case the rename is rolled back instead of leaving the file and the database at two different locations.

Path Drift Reconcile was added because a normal root/path resolver can translate Docker or media-server prefixes, but it cannot know that `01-01 - Song.flac` in the database is now called `01 - Song.flac` on disk. The tool proposes a correction only when there is exactly one safe candidate in the same directory, it never takes a path that another file row already owns and it refuses ambiguous matches.. approving the finding only repoints the index and never renames, moves or deletes the real file.

### Track Number Repair

Track Number Repair was migrated from the legacy album/track rows to Library V2 editions and concrete file subjects. This was important because Library V2 separates a release group from its actual editions, while the old tool only had the flatter album view.

The first native version still combined every track of the complete release group. With a normal and a deluxe edition this could calculate something like track `4/28` for a 12-track edition and then write this wrong value into every affected file and rename it too..

Now every file is checked against its own concrete edition and its own release-track position. If the edition cannot be identified clearly, the job creates no finding instead of guessing. For a tool that can rewrite and rename a complete album, no result is much safer than a result that only looks plausible.

### Dead File Cleaner, Orphan File Detector and Expired Download Cleaner

These jobs needed the new Library V2 file lifecycle because a path alone is no longer the complete state of a file.

Dead File Cleaner now only works on files that passed through the Library V2 missing lifecycle and reached `missing_confirmed`. A file is first `missing_suspected`, and an unavailable storage root, NAS or Docker mapping is never allowed to confirm the miss. The deleting fixes also check root health again before changing the catalogue.

It also checks if another active file of the same track still exists.. removing one dead MP3 must not demonitor a track that still has a valid FLAC and must not create a new download for something the user still owns.

Orphan File Detector now checks the legacy index, Library V2 and the filesystem before it calls something an orphan. It still uses the path instead of inventing an entity ID, because a real orphan has no catalogue entity by definition. Path Drift Reconcile sits before destructive cleanup so an externally renamed file can be reconnected instead of appearing once as missing and once as orphaned.

Expired Download Cleaner stays filesystem-based, but before deleting anything it must be able to resolve and record the affected Library V2 subjects. The deletion then goes through the central file lifecycle so the catalogue, Wanted state and history cannot silently disagree with the disk.

### AcoustID Scanner and verification

AcoustID Scanner was migrated to active Library V2 file subjects and now persists its observation into the Library V2 verification state. The audio file carried the real `SOULSYNC_VERIFICATION` marker, but the first native scan inside this branch read this value and then dropped it before writing the snapshot.. this is why many files appeared as never verified even when the proof was inside the file.

When an AcoustID fix moves a file onto a corrected track identity, the old expected track can become fileless. Wanted is now recalculated after this change, otherwise the album could still look complete while one expected track had no file anymore.

The UI also got separate Check and Verification information so the user can see the difference between verified, human verified, skipped and not scanned instead of guessing from the Quality field.

### Fake Lossless, Corrupt File and Short Preview tools

These tools were migrated from legacy tracks to active Library V2 files and filesystem subjects.

Fake Lossless intentionally stays report-only because i dont want a destructive automatic action based only on this analysis.. but its finding now carries the real Library V2 file identity. The first native version accidentally stored a track ID under `entity_type=file`; because track and file numbers can overlap, it could point at a completely unrelated file.

Corrupt File and Short Preview use the shared path resolver and root guards so they inspect the actual mounted file. If a destructive action is approved afterwards, it runs through the same Library V2 delete/Wanted lifecycle instead of only removing a legacy path.

### ReplayGain Filler, Lyrics Filler and Lossy Converter

These tools were migrated to `lib2_track_files` and had to learn the Library V2 multi-file model. A track can have a FLAC, an MP3 or another retained copy at the same time, and exactly one of them is only the primary file.. primary does not mean that the other files may be ignored.

ReplayGain, Lyrics and Re-tag now update every writable file of the selected track instead of changing only the primary and still reporting the whole track as successful. ReplayGain analyses every file separately because a FLAC and an MP3 do not automatically have the same measured gain.

Lossy Converter registers the converted output as a new Library V2 file. If the original lossless file is kept, it also remains attached as a second file of the same track; if it is replaced, the removal goes through the central lifecycle. Before this convergence work a retained original could later appear as an orphan, while Quality evaluation only saw the lossy copy.

### Album Tag Consistency, Discography, Quality and monitoring jobs

Album Tag Consistency was migrated to the native Album -> Track -> File relations and now rescans the exact affected files after applying a correction.

The old Duplicate Detector logic was not copied into another V2 finding engine.. Library V2 has its own Artist/Album/Edition dedup and reconciliation because it has stronger provider identities and knows the difference between a release group and an edition.

Quality Upgrade uses the same app-wide Quality Profile evaluator as Search and Import. It was fixed to resolve the effective profile live instead of trusting an old copied profile ID on the track.

Monitored Discography Refresh uses the native discography expansion, monitor rules and Wanted projection. Monitoring List Reconcile connects Library V2 monitoring to the existing Watchlist and Wishlist in both directions and skips an old Outbox operation if a newer user decision already replaced it.

Cache Maintenance and Empty Folder Cleaner intentionally remain catalogue-free.. they operate on cache files or empty directories and dont need a fake artist/album/track identity only to look “native”.

So the reason for touching the tools was not to change what SoulSyncs tools are.. it was to migrate them from the only library that existed before this branch into the new catalogue, and to make sure that after a fix the disk, Library V2, monitoring and history all tell the same story.

## The other work since the last update

Besides the tools i also fixed and improved Interactive Search, artwork handling, multi-file imports and replacements, monitoring and Wishlist mirroring, live Quality Profile resolution, download-client adoption, async handling and alot of Library V2 UI details.. I also completed the new artist discovery flow described above and made the newest tracklists resolve in the background without locking the complete SQLite database.

I made a final broad bug run over the complete Library V2 flow and fixed every confirmed finding from it. The Library V2 backend passed repeatedly with more than 1,000 tests, the frontend passed more than 300 tests including its checks and production build, and important migration/repair flows were also run against a backup of my real database and real audio files..

There can of course always be another special provider, client, restart or path-mapping edge case.. but after all these runs im very very confident that the base and the dangerous state transitions are now correct, and im fully committed to continue fixing and improving this after merge too..

its been hard to hold on with your pace.. so i think after you give this one final try and some feedback it would be ready to merge.. i dont really want to drag this much longer because with how fast dev is moving it gets harder and harder for me to keep everything compatible and mergeable..

so tell me what you think and what you still see as a blocker.. and then we can move forward..

nick2000713:
@Nezreka hey, before i clean anything up i'd rather ask you.

  when i started this, the artist page was still the old vanilla one and part of
  the library i'm replacing here. so i rebuilt that behaviour inside Library V2 —
  search hit lands on the artist, discography, filters, rich header, the lot.
  while i was doing that you kept going and moved the whole webui to react,
  artist-detail included. so the thing i built against isn't the same thing
  anymore. i only saw how much had changed when i synced onto dev.
  
  right now `/artist-detail/$source/$id` redirects into Library V2. it sits in the
  route and not in the callers, because all six entry points build that url
  through `buildArtistDetailPath` (search, global search, media player, playlist
  sync, similar artists, stats) — one redirect covers them all and old links and
  history keep working. but it also means **your artist page is currently
  unreachable**. it's still in the tree, one file revert away.
  
  two ways i can go, and it's your call:
  
  1. you're fine with V2 owning the artist page → i delete
     `webui/src/routes/artist-detail/` (~7k lines plus its tests), so we're not
     shipping a page nobody can open
  2. you want to keep yours → i limit the redirect to artists that are **not in
     the library yet**, and your page keeps everything else
  
  i really don't mind either way. you know what you have planned for that page and
  i don't.
  
  what V2 covers, so you don't have to dig:

  - gap-fill (#1067/#1071), completion overlays, rich header with
    listeners/plays/top tracks, discography filters, table ↔ card toggle — the
    header and the cards even reuse your css classes, so it looks like what people
    already know
  - what Enhanced does (per-track metadata editing) — V2 has its own inline and
    bulk editor, plus quality profiles, monitoring, wanted, history, tag writing,
    lyrics and replaygain on the same page
  
  what it does not have, so you know exactly what you'd be giving up:

  - the DB Record inspector — no equivalent, would be a small port
  - the enrichment coverage rings / metadata sources panel — V2 shows match chips
    instead. that was a deliberate call early on, not an oversight, happy to
    revisit
  
  could you give it a run before deciding? search an artist you don't own → you
  land in the V2 discovery view → "Bookmark" → it materialises and you're on the
  full V2 artist page. then the same for an artist you already have.
  
  and honestly, this is exactly why i'd like to get this PR closed at some point.
  every week it stays open the base moves and i end up rebuilding against
  something that changed again. happy to split it up or do whatever makes it
  quickest for you to review.
  
  not deleting anything until you've had a look.

nick2000713:
Hey @Nezreka,

Coming back to what I said further up — that some of this doesn't belong in this PR and could be split out. I've now gone through it properly instead of just claiming it, so here's a concrete proposal.

**First, one correction I owe you**, because the description at the top of this PR is out of date and I don't want you reviewing against a wrong picture:

The `features.library_v2` flag is effectively dead in the current branch. It always returns true. That happened when I migrated the repair/maintenance tools: a native job with the flag off doesn't fall back to the old tables, it reports a clean, empty scope — which is worse than not shipping it, because nothing tells you the tool silently stopped working. So I made the cutover non-reversible rather than ship a switch that quietly breaks 30-odd tools. It's the right call technically, but it means this is **not** the opt-in feature the description promises anymore, and that's a much bigger thing to merge in one go. Which is really why I want to split it.

---

**The split I'd suggest:**

**1. Bugfixes that already affect `dev`** — this is #1121, already open and separate. ~2.3k lines, no new tables, no UI, nothing to opt into. Things I hit by accident while building: the Prowlarr search timeout being a hard 15s constant no setting could raise, a failed Prowlarr search being indistinguishable from zero hits, category 3060 missing so non-Latin-script releases were unfindable over Usenet, `safe_move_file` deleting the source because the destination "existed" (it's the library file, it always exists on a replace), the auto-import approval path, and #1109. Plus an event-loop fix where a lost selector wakeup on Python 3.14.6 hangs a subsystem permanently with nothing in the logs.

None of it is Library v2's to carry, and it's all live on `dev` right now.

**2. Schema + ingest + the repair/maintenance tool migration, together.** This is the piece I'd most like you to look at first, because it's the piece that explains why this PR got so big in the first place. It's also the only part of the whole thing that makes the codebase **smaller** — it deletes 10 legacy jobs and consolidates the rest, roughly −9k lines against +26k, across 83 files.

What's in it: the `lib2_*` tables, the importer, the startup bootstrap that migrates an existing library into them, and the 33 tools rewired to read from that catalogue instead of each keeping its own copy of path resolution and its own assumptions about where a track lives.

The three of those have to travel together, and it's worth saying why: the native jobs enumerate their subjects from the catalogue, with no fallback. Ship them without the bootstrap that populates it and every single one reports an empty scope on upgrade — silently. With the bootstrap in the same PR, an upgrade builds the catalogue in the background and the tools work, which is exactly what would happen if this whole PR merged as one anyway. So it's not a new risk, it's the same migration in a smaller box.

I checked the boundary before proposing this rather than assuming it: the repair suite reaches into 42 of the 64 v2 modules and **zero** API routes, zero acquisition code, zero frontend. The Tools page needs no changes at all. So this really does cut cleanly.

**3. The rest** — the v2 API, the UI, acquisition, and the artist-page cutover. Still substantial, but by then it's the part that genuinely *is* Library v2, sitting on a data model you've already reviewed and a tool suite that's already migrated and has had real-world exposure.

---

**Two other things I want to do regardless of whether we split:**

- The internal working notes I've been keeping in `docs/` (roughly 11k lines of my own issue journals and status logs) shouldn't be in this PR at all. Those are my notes, not project documentation. Dropping them is free.
- `library-v2-page.tsx` is a 10.6k-line single file. That's not reviewable by anyone, including me. Splitting it into modules doesn't reduce the line count, but it's the difference between "can be reviewed" and "can't".
- When step 3 lands, the old library React layer goes with it — `-library.api.ts`, `-library.helpers.ts`, `-library.types.ts`, `-library.live.ts`, `-library.watch-all.ts`, `-library.export.ts` and the three components under `-ui/`. That's another ~1.9k lines. Nothing outside `routes/library/` imports any of it and the components have no importers at all, so it's already dead weight on the branch today — I just haven't swept it yet.

---

Does 1 → 2 → 3 work for you? I'd keep this PR as the tracking one until the last piece lands. And if you'd rather just review it as one thing after all, that's fine too — tell me and I'll stop reorganising and go back to fixing what you find.

Nezreka:
Sorry im here now to review this. The react migration ate weeks of my time. I'll get back to you soon

Nezreka:
I am so impressed with this new library page. Let me play around with it some more but this feels rock solid and so much better, so thank you.

nick2000713:
Thank you, im realy happy you like it and that it already feels rock solid.. take all the time you need to play around with it xD
 
i still see alot of things i would like to optimize and some little features i want to add.. but i intentionally didnt put even more into this PR because honestly its already way too large even for me now xD

i would rather first get this base merged and make sure it is realy rock solid with some actual use.. and then add those things afterwards in smaller follow-up PRs, one step at a time..

but im fully committed to keep working on it and make it even better, so take your time and just tell me everything you find.. i will take care of it...

Nezreka:
ok here's everything i found.

tested on my library. 4,979 artists, 69,296 albums, 307,885 tracks, 9gb db.

## the migration

it got stuck. sat at "5/7 · 71%" for 9 minutes with no log output. every enrichment
worker was throwing "database is locked" the whole time, plus the automation engine.
then a config save failed too, "Config DB save failed after 6 attempts".

i restarted, which made it worse. second start never got past db init. last log line
12:05:56 then nothing for 30 minutes. it was still writing, i could only tell by
watching the wal:

    12:07  wal=85mb
    12:14  wal=106mb   (~4mb/min)
    12:35  wal=135mb   (~1.4mb/min)

slowing down, not speeding up. main db never moved off 9,652mb. i gave up at 30 min.

what i need before merging:

- progress logging on a timer. "41,000/307,885" every 30s or maybe faster. right now there's no way to
  tell if it's working or hung
- pause the enrichment workers while it runs so they're not fighting it for the writer
- keep it off the startup path so the server comes up first
- checkpoint the wal so it stops slowing down

## enrichment

this is the main thing. i don't want to lose any enrichment functionality or data in the
move to v2. v2 should get filled the same way v1 does now. every artist and album,
artwork, genres, bios, provider ids, all twelve workers.

v1 already does this, so it should be doable.

three things i found that make me think it isn't happening yet:

1. resync_entity_from_legacy doesn't look wired up. the docstring says it runs after
   _run_single_enrichment in web_server.py but i only see it in its own module and its
   tests. if that's right the workers write legacy and nothing brings it into lib2
2. artists that came from the old library get all twelve workers. artists created in
   lib2 (featured credits, wishlist, discography) have no legacy row so they only get
   native_enrich, which does provider id, artwork and genres. both show as matched so
   you can't tell them apart
3. /api/library/artists still reads the legacy tables. so metadata edits and enrichment
   done in the new ui don't show up in the api

## tables

lib2_track_artists and lib2_track_files have to exist, legacy tracks only has one
artist_id and one file_path, no way around that.

but lib2_artists, lib2_albums and lib2_tracks look like copies. i went through
lib2_artists column by column and they could all have been columns on the existing
artists table.

is that the end state or is it temporary? i'd rather end up with one library, everything
reading and writing lib2, and legacy either gone or read only. what i want to avoid is
both sticking around and disagreeing with each other.

## small one

mbid_mismatch_detector is in RETIRED_JOB_IDS but not in PRESERVED_RETIRED_FINDING_IDS so
its findings get pruned at worker start. doesn't affect me, i have none. your
library_reorganize comment right above covers the same hazard so i assume this one's
deliberate, just checking.

the page itself is great. schema.py reasoning is right, i checked it against the old
schema. just want the migration sorted and the enrichment not to regress.

nick2000713:
thx alot for testing this on such a huge library and writing everything down soclearly.. i will dig into all of this now.

and sry about the migration.. i realy thought i had solved it for big librariesalready xD, but yours is on a completely different scale and clearly exposed some things my test DB did not..

i will get back to you once i worked through everything.. thx again, this is exactly the kind of real test i needed.

nick2000713:
@Nezreka, i worked through all your points now and honestly this became alot rougher than i expected xD.. you were right to keep asking which database we actually want to maintain in the future. keeping legacy and lib2 alive as two equal libraries makes no sense.. every import, metadata edit, enrichment result and repair job would need to be synced forever, and sooner or later both databases would disagree again.

so i stopped treating lib2 as an optional second library. lib2 is now the main Library and the source of truth for normal runtime. the old tables are only kept for the first upgrade import and as a temporary safety/rollback boundary. i want to remove them in a later update, but only after this survived another real test on your 9gb database.

i also thought alot about why not just put everything back into legacy.. but the old model simply does not fit anymore. one legacy track has one `artist_id` and one `file_path`. it cannot properly represent multiple credited artists, multiple files/qualities for one recording, single↔album relationships or mappings from several media servers. to keep legacy i would basically have to rebuild its schema and rewrite every query and worker anyway, while still carrying all those old assumptions. lib2 already models these things properly, so one controlled migration to lib2 is much safer than permanent translation between two libraries.

about the problems you found:

- **migration stuck at `5/7 · 71%`:** the server now starts first and the large migration continues in the background. it logs the real stage and counter every 30 seconds, including when the counter did not move, so it should finally be possible to see the difference between slow and stuck
- **all the `database is locked` errors:** enrichment workers and the automation engine are paused while the migration owns the writer, then only the workers that were running before are resumed. the large passes commit in batches instead of one huge transaction
- **the WAL growing and getting slower:** the migration checkpoints the WAL between batches, so it should not just grow from 85mb to 135mb while throughput keeps dropping
- **restart making it worse:** progress is stored with a real resume checkpoint. after a restart it should continue the matching stage instead of blocking database init or starting the complete catalogue walk again
- **no enrichment in lib2:** your reading of `resync_entity_from_legacy` was right at the time.. it existed but was not reliably wired into all worker paths. i moved the provider enrichment to lib2 instead of depending on that bridge forever. artwork, genres, bios, provider ids and provider payloads now end up on the lib2 entity
- **native-only artists getting less enrichment:** migrated artists and artists created from featured credits, wishlist/watchlist or discography now use the same native lib2 enrichment path. not having a legacy row should no longer create a second class of "matched" artist with less data
- **`/api/library/artists` reading legacy:** the active Library API/UI now read and write lib2. the old compatibility response is projected from lib2 too, so metadata edits and enrichment should not disappear depending on which endpoint reads them
- **which tables are the final ones:** `lib2_track_artists` and `lib2_track_files` stay for exactly the reason you mentioned. and `lib2_artists`, `lib2_albums` and `lib2_tracks` are now the maintained catalogue, not temporary copies beside another active library
- **`mbid_mismatch_detector`:** it is now an active lib2 job, not a retired job whose findings get deleted at worker startup

i added a ratchet test for this too.. normal production Library code is not allowed to add new reads or writes to the legacy artist/album/track tables. the one-time upgrade reader is counted separately, so we cannot quietly start depending on legacy again later.

the media-server behaviour is also clearer now:

- imports always create/update the local lib2 row and do not wait for Plex/Jellyfin/Navidrome to be online
- a media server does not create the Library anymore. after a successful scan it only adds its own recognition mapping to the existing entity
- the Library shows a small server/check badge when a server recognised the artist/album/track
- a failed or interrupted refresh keeps the last good mappings. they are only replaced after the server was read successfully
- SoulSync DB metadata is not silently pushed into Plex/Jellyfin. the server benefits when SoulSync writes the actual file tags and the server scans the file again

i merged the newest dev too and checked the conflicts one by one so we keep the fixes from there. manual artwork now has a real lock in lib2 and survives provider/server refreshes, the new thumbnail cache is connected to the active lib2 page, genre enrichment scans lib2, reorganize keeps repair findings on the new path and i fixed a crash in the automatic native enrichment sweep that only showed up when starting the real app.

the final merge is `f7496e8f0` and includes dev through `bbf21fd18`. no rebase, no force push and no replaced history. the PR is clean/mergeable now and none of our guide/issue/status markdown files are part of its diff.

if you have the time i would realy like you to test these painful cases again:

1. upgrade the same pre-lib2 database and check that the server/UI comes up while migration is still running
2. watch the new counters and WAL through the huge track stage
3. restart once during migration and check that it resumes without duplicate or missing rows
4. compare enrichment on one migrated artist and one native-only wishlist/featured artist
5. edit metadata and artwork, then run enrichment/refresh and confirm the values still stay
6. import while the media server is offline, then scan it later and check that the badge appears without a duplicate Library row
7. interrupt a media-server refresh and confirm the previous mappings/badges remain
8. have a look at reorganize findings, genre enrichment and `mbid_mismatch_detector`

i ran everything before pushing:

- backend: `15307 passed, 3 skipped, 2 deselected`
- webui: `325 files / 6724 tests passed`
- browser e2e: `30 passed, 8 skipped`
- production build, ruff, format and type/lint checks passed
- lib2 legacy/native ratchet: `17 passed`

the tests are clean now, but your database is still the test that matters most because it is on a completely different scale than mine.. so pls tell me anything that looks stuck, missing, duplicated or just behaves differently. even one weird provider or badge is useful and i will dig into it..

thx again for pushing on this.. the rewrite was rough, but your question about which database we actually maintain was the right one and the architecture is alot clearer now..

nick2000713:
@Nezreka Quick update since the last message:

  - Ported Reassign to a different artist to Library V2 and tested it live in the browser.
  - Kept the branch in sync with dev — just merged your latest 7 commits (concurrent-downloads fix, post-processing queue
  fix, preferred-version matching, Navidrome path resolution, per-item quality-profile filter, radio-edit detection). All
  conflicts resolved, full test suite green after the merge.
  - Kept fixing real bugs found by actually using the app against my own library: permanent file deletion, AcoustID scan
  status not updating live, a few missing-count/history display issues.

No urgency on a reply, just wanted to keep you posted that I'm tracking what you implement upstream, and making sure it lands in Library V2 too...

nick2000713:
@Nezreka synced with dev again (up to 250f7f90b), conflicts all resolved on my side.

also ported the new work over to library v2 here, metasync export and the expired cleaner queries and soul_id_path. i'll keep doing that with every sync, it just means the PR keeps growing a bit each round.

i'll keep doing it, but the PR gets bigger the longer this goes on.. splitting it isn't really an option anymore either, the cutover is one piece since i ported every worker...da

nick2000713:
@Nezreka quick update since my last message (21 Aug), just pushed.

synced 5 times since then (3.2.3 to 3.3.2), all conflicts resolved on my side, suite green every time. thanks for merging #1169/#1172/#1173/#1182/#1185/#1190/#1191 too, good to see that stuff land on dev directly instead of only living in here.

**biggest find:** guests were inheriting the album artist's identity. 47 rows across 6 groups had someone else's id.. 12 artists were walking around with Major Lazer's photo, 10 with Sawano Hiroyuki's, their cached photos were byte identical to his. root cause was one detail in the tracklist adapter, it only kept credit names, so a guest became an id less row, and the enricher then guessed an identity for that row from any album it was credited on, which is always the album's PRIMARY artist. fixed that, also cleaned out 82 empty ghost artists that had no library and no reason to exist.

**a handful of catalogue bugs i only found by actually running this against my own library and a production db dump:**
- 21 recordings that exist on both an album and a single were treated as two unrelated files instead of one.. 2 of them had already been silently re downloaded and now sit on disk twice
- 368 releases were flagged complete on zero real evidence, the check was comparing a number to itself, always true
- 373 out of 373 v2 wishlist rows had no artwork at all
- corrupt file findings could never actually be fixed, both fix handlers required a track id and that finding type never has one

**reorganize and retag got rebuilt.** i already opened this against dev too, #1186, same diagnosis, reorganize was secretly running a file you already own back through the whole download acceptance pipeline. that's what quarantined a correct Sawano file, AcoustID read the kanji name off the file while the catalogue held the romaji one. #1186 has the long version, nine separate opt out flags that all said the same thing in different words.

first, how a song actually gets in, because there are three ways and only one of them is a download. manual grab is you picking a specific result. automatic search picks for you, and if it finds nothing or the download dies the track goes on the wishlist, which retries it later, so the wishlist sits behind those two rather than beside them. and then there's Import, where you drop files into staging yourself. that one's different because nothing told it what the files are, so it has to identify them first, and if it isn't confident enough it parks them for you to confirm instead of guessing:

```mermaid
flowchart LR
  MG(["manual grab"]) --> DL["download"]
  AS(["automatic search"]) --> DL
  DL -.->|"failed or\nnothing found"| WL[("wishlist")] -.->|"retries later"| DL
  IM(["import: you drop\nfiles in staging"]) --> ID["identify by tags,\nfilename, AcoustID"]
  ID -.->|"not confident\nenough"| RV[("parked for you\nto confirm")] --> CHK
  ID --> CHK
  DL --> CHK
  CHK["checks: integrity,\nsilence, AcoustID"] -.->|"fails one"| Q["quarantine"]
  CHK --> TAG["write tags"] --> P["build path"] --> MV["move into library"] --> CAT[("catalogue")]
```

from the checks onward it's the exact same code for all three, and the catalogue row is what proves it landed. a file sitting on disk without that row does not count as imported, that's an explicit guard now rather than something you hope happened.

after that, four things can touch that row, and each only does its own job instead of all of it happening inside reorganize:

| step | reads | writes |
|---|---|---|
| you, editing a field | nothing, it's your call | that one field, and marks it as yours |
| manual match | your pick of the correct release | just the provider id on the row |
| refresh | that id, then the provider | title, track numbers etc, skips any field marked yours |
| re tag | the row | the file's tags, same skip rule |
| reorganize | the row | the file's path, moves it |

**the Re-tag job under Tools** is the scheduled version of that fourth row, and it is scan only, it never writes anything by itself. it walks the library, compares each file's embedded tags against what the catalogue holds, and raises one finding per track that has drifted, "Tags out of date: Artist - Title", with the full field diff on it. applying a finding runs the same engine the dialog uses, so what the finding promised is what actually gets written, which was not true before. off by default, weekly when you turn it on, and scoped, so running it for one artist can't produce library wide findings. a file it can't even read counts as an error instead of becoming a finding, because a finding promises a fix.

the bit i actually want to show you is what that first table row does to the rest. you can edit any track or album field directly in the library, title, genre, style, whatever. the catalogue then remembers not just the value but that it's yours, per field. same catalogue box as in the diagram above:

```mermaid
flowchart LR
  U(["you edit a field"]) -->|"stores it and\nmarks it yours"| CAT[("catalogue")]
  RFR(["refresh, from provider"]) -.->|"skips fields\nmarked yours"| CAT
  CAT --> RT["re tag preview\nor Tools finding"]
  RT -->|"field isn't yours"| W1["writes it,\nno question"]
  RT -->|"yours, and the provider\nwants something else"| W2["shows both,\nyou pick per field"]
```

a real one from testing, three values on one row instead of two: the file has `Vogel Im Kafig`, your edit says `Vogel im Käfig`, the provider wants `Vogel im Käfig (OST)`. yours wins by default, one click flips it, and a bulk apply counts them for you first, "23 findings, 4 of them yours", before it touches anything.

and honestly the nicer part is that the old failure mode mostly can't happen anymore. it's not that i added guards against reorganize renaming your files back, it's that re tag and reorganize both read the catalogue now, and the catalogue is where your edit lives. neither of them reaches past it to a provider, so there's nothing left to disagree with.

that's most of what's been sitting here. i think this closes out the backlog i owed you, planning to slow down on new changes for a bit and let this settle.

wishx:
I have 230,915 tracks in my library, and I'm ready to test this tout de suite.

nick2000713:
@wishx  nice, 230k tracks is a lot more than i have here, so you'd be testing it on something i can't reproduce myself. that's exactly what the importer needs, it has to hold up on other people's libraries, not just mine.

one thing first though: back up the whole data dir with the container stopped, not just the .db. first start migrates your existing library into the new catalogue, it's resumable and can't run twice, but it's still the one step that touches data you already have.

if anything breaks or drags at that size, post it here.. numbers help. and im very curious on what you think..

wishx:
I'm reaaaaally loathe to go poking around in .json files, espacially config files. Would it be possible to add a toggle to Experimental in Advanved Settings or something? Like, toggle it on, restart the container, and Bob's your uncle?

image

nick2000713:
@wishx good news, that section is outdated and thats on me.. the json step is gone.
  
theres no features.library_v2 switch anymore. lib2 is the catalogue now, not an opt-in second library, so its simply on. if the key is still sitting in your config it gets read once and ignored, nothing to toggle.
so its just: back up the data dir with the container stopped, then start. first start does the migration, its resumable and cant run twice. thats everything.

ill rewrite the "how to try it" block in the issue so the next person doesnt walk into this. still very curious how the importer holds up at 230k..

PR #1076, opened 2026-07-25
Quality profiles foundation

nick2000713:
# Quality Profiles Foundation

## Why this PR exists

Quality Profiles already exist in Settings, but today they're mostly decorative: the only place you could actually assign a named profile to something was auto-import. Everything else — Watchlist artists, mirrored playlists, manual Search/Discover/Library downloads, Auto-Sync — just used one global quality setting, regardless of how many profiles you'd defined. You could build five profiles and only ever use one of them.

This PR is what turns Quality Profiles from a settings-page curiosity into something that actually governs downloads: **you can now say, at the point a download happens, which profile it has to satisfy** — per artist, per playlist, or per individual action — instead of that decision always falling back to whatever the single global default is. There was previously no way to say "this artist should always be lossless" or "this playlist is fine at 320kbps MP3"; changing the global default silently changed what *every* automation downloaded from that point on, including things you didn't mean to touch.

This PR fixes that at the root: it gives **individual Watchlist artists** and **individual mirrored playlists** their own durable, named Quality Profile, instead of all sharing one global value. That assignment is then honored everywhere a download can originate — manual search, Discover, Library, Auto-Sync, playlist Sync, Download Missing, and retries of anything that failed. That's the missing half of the feature; this PR is what unlocks the rest of what Quality Profiles were built to do.

It's built to be useful on two timelines at once:

- **Right now**, on its own, with nothing else needed: you can already set "Artist X always downloads FLAC" or "this playlist stays MP3 320" and have every future scan, sync and retry respect it, without touching a second, unrelated part of your library.
- **Later**, as the piece Library v2 needs and doesn't have to reinvent: Library v2's whole premise is per-artist/per-release management, and it needs a quality assignment that lives outside of it, survives it being cancelled, and that it can just call into rather than owning itself. This branch is deliberately Library-v2-*aware* — the API contract is shaped for what Library v2 will need — but Library-v2-*independent*: it contains no Library v2 schema, imports, or components, and works exactly the same if Library v2 never ships.

## How it works

Every Watchlist artist and every mirrored playlist gets exactly one `quality_profile_id`, pointing at a named Quality Profile (the same profiles already configurable in Settings). Nothing about the JSON quality rules themselves is copied around — every consumer resolves the *named* profile live at the moment it's needed, so editing a profile's rules in Settings immediately changes the behavior of everything assigned to it, without a migration or a re-scan.

From there, the assignment just flows downstream instead of being re-decided at each step:

- A Watchlist scan writes the artist's profile onto every Wishlist row it creates.
- Auto-Sync stamps the mirror's profile onto every track it generates, on every run.
- Manual Sync and Download Missing resolve the same mirror assignment before they start.
- If a download fails and falls back to the Wishlist, or gets requeued, or gets re-added from sync history, the *original* requested profile travels with it — a retry never silently reverts to whatever the global default is by then.
- Manual, one-off downloads (Search, Discover, Library) get a per-action Quality Profile selector right in the existing "Add to Wishlist" / "Download Missing" dialogs, so you can override the profile for just that click without touching the artist's or playlist's persistent assignment.

Nothing is left ambiguous about *whose* choice wins: an explicit choice on a single action beats the playlist's saved assignment, which beats the app-wide default — always in that order, and it's the same order on every endpoint.

## What changed, concretely

**Data model**
- `watchlist_artists.quality_profile_id` and `mirrored_playlists.quality_profile_id`: nullable during the schema transition, then backfilled at startup to a Quality Profile generated from the current global config, so nothing changes behavior on upgrade.
- `sync_history` gains `profile_id` and `quality_profile_id` (existing rows read as `NULL` = legacy/admin).
- Deleting a Quality Profile that's still assigned somewhere repoints those artists/playlists to the remaining default instead of leaving a dangling reference. A `config.json` reference (`auto_import.quality_profile_id`) can't be cleaned up atomically with the DB delete, so a small idempotent startup reconciler retries that cleanup until it succeeds.

**Native API contract**
- `core/watchlist_sources.py` is now the single provider → id-column registry (Spotify, iTunes, Deezer, Discogs, MusicBrainz, Amazon + aliases), replacing an old heuristic that guessed the provider from the shape of the id ("all digits means iTunes") and mis-filed anything else. An unrecognised `source` is now a `400` instead of a silent misfile.
- Every optional field across the Watchlist/Wishlist/mirror endpoints now has three distinct, explicit states: absent/null → keep the existing value (or default, for a new row); valid → use it; invalid/unknown → `400`. It never silently falls back to a default when a real value was sent and rejected.
- Mirror endpoints are ownership-scoped by `(mirror_id, active_profile_id)`: a mirror belonging to another SoulSync profile now reads as `404` everywhere — read, rename, preferences, pipeline run, discovery clear, export, delete, resolve — closing a cross-profile access hole where a guessable id could expose another profile's playlist and its assigned quality.
- The preferences PATCH validates the Quality Profile id and `organize_by_playlist` together before writing either, so a rejected profile id can't leave a half-applied change behind.

**UI**
- Persisted selectors (Watchlist settings, playlist detail, mirrored-playlist/Auto-Sync cards) never show a value the server doesn't actually hold — "Mirror playlist first", "Quality Profiles unavailable", and "assigned profile was deleted" are distinct, visibly inert states rather than silently defaulting to whatever option is first in the list.
- A failed save rolls the control back to the last value the server actually confirmed, instead of leaving an unsaved selection looking active.
- Real styling on both selector variants (persisted vs. per-action), matching the rest of the app instead of the unstyled native control.

## Review history

This branch went through two internal review rounds before being opened here — 30 findings total across both rounds (validation gaps, the cross-profile mirror-access hole, atomicity, silent-fallback behavior, UI state honesty, stale-cache/race conditions in the frontend). All 30 are fixed and covered by regression tests; nothing is being carried forward as a known issue.

## Test coverage

New/updated tests cover: default-profile inheritance for a new Watchlist artist, explicit-assignment overwrite, profile-deletion repoint, wishlist-intent overwrite on repeated authoritative writes, a mirror's assignment surviving a provider refresh, non-Spotify source resolution, the Auto-Sync fingerprint reacting to a profile-only change, failed-download preservation, foreign-mirror `404`s, the atomic preferences PATCH, per-profile sync history and retry, and the full manual-acquisition precedence order.

```
tests/acquisition/test_quality_profile_contract.py
tests/artists/test_quality.py
tests/automation/test_handlers_playlist.py
tests/database/test_mirrored_playlist_preferences.py
tests/database/test_mirrored_playlist_scoping.py
tests/database/test_mirrored_playlists.py
tests/database/test_sync_history_quality_profile.py
tests/database/test_sync_history_scoping.py
tests/playlists/test_explorer.py
tests/playlists/test_explorer_scoping.py
tests/quality/test_batch_quality_profile_resolution.py
tests/quality/test_native_api_quality_profile_contract.py
tests/quality/test_stale_config_reconcile.py
tests/quality/test_wishlist_add_outcome.py
tests/quality/test_wishlist_quality_columns.py
tests/watchlist/test_batch_add_reporting.py
tests/watchlist/test_partial_config_update.py
tests/watchlist/test_provider_neutral_contract.py
tests/watchlist/test_quality_profiles.py
tests/wishlist/test_service.py
```

All changed Python and JavaScript files pass syntax checks; `webui/static/*.js` is checked in script mode rather than `node --check` (the directory is marked ESM in `webui/package.json`, which rejects pre-existing duplicate top-level function declarations that are legal in the classic scripts the browser actually loads).

## Non-goals

- Finishing or shipping the Library v2 Playlist tab.
- Defining Track > Album > Artist > Playlist conflict precedence in Library v2.
- Rewriting provider-specific playlist discovery.
- Migrating already-completed media files when an assignment changes retroactively.

## Notes

- Internal review/audit notes for this branch were kept out of this PR's diff (working documents for the review process, not user-facing).
- Base: `upstream/dev` at the time this branch was cut.


nick2000713:
Since this is the piece Library v2 needs and was deliberately built to be Library-v2-aware, I'm pausing Library v2
  development until this merges, then rebasing on top of it and reconciling the profile handling that Library v2
  already has in progress.

PR #1081, opened 2026-07-25
fix(tools): Comma Artist Splitter - scan files and properly split multi-artist metadata

Lenochxd:
## Problem
The Comma Artist Splitter job couldn't detect most multi-artist tracks in libraries:
- Scanned only the database (which stores only the first artist most of the time)
- Real multi-artist metadata exists only in file tags (e.g., "Artist A, Artist B")
- Detection rate was extremely low (~22 finds in 10k+ tracks for me)
- Fix handler couldn't apply changes due to artist_id lookup mismatch

## Solution
Complete rewrite to scan actual file metadata instead of database:

1. **Scan Layer:**
   - Iterate through all tracks and read file metadata directly
   - Extract ARTIST tags from all formats (ID3, MP4, Vorbis)
   - Support multiple separators: comma (,), semicolon (;), ampersand (&), slash (/)
   - Normalize all separators for consistent API matching

2. **Fix Layer:**
   - Store scanned file paths in finding details for the fix handler
   - Use file list from finding instead of database artist_id lookup
   - Properly write split artists to all detected files

## Changes
- Add `_get_artist_tag()` helper for format-agnostic tag extraction
- Rewrite `scan()` to process files instead of database artists
- Update `_fix_comma_artist_split()` to use finding's file list
- Rename `split_comma_parts()` to `split_artist_parts()` for clarity
- Update `estimate_scope()` to count tracks instead of DB artists
- Update tests for multi-separator support

## Impact
- Dramatically increases detection
- Properly handles diverse metadata formats and separator conventions
- Now detects 'Artist A, Artist B', 'Artist & Guest', 'A; B; C', 'X / Y' etc.
- Splits with semicolon display format + proper multi-value tags (Picard convention)


Nezreka:
you're right about the core problem, the multi-artist info lives in the file tags and the old name scan was blind to it. scanning tags is the right call. couple things before i can merge though, mostly cause this writes to people's files:

- **tests are red.** only the split helper test got updated. the scan/fix tests still use fake paths, so `test_flags_split_when_parts_in_library_and_api_says_not_real` fails (scan skips the file since it doesn't exist), and the new tag-reading path has no real coverage. need them rewritten to write actual tagged files and assert the new shape.
- **dedup is broken.** findings go in with `entity_id=None` and `file_path=None`, so neither branch of the dedup check matches, every scan re-adds the same findings and fixed/dismissed ones come back. give the finding a stable entity_id (combined name works).
- **& and / worry me.** they cover a lot of real band names (simon & garfunkel, ac/dc, hall & oates). the guards should catch most but a wrong split corrupts tags and none of it's tested, want real cases proving those don't split.

smaller stuff: scan now reads tags for ~3000 files a run so it's way heavier, the rate-limit pause moved so api lookups fire back to back, and `all_files` only has files seen in this run's window so ones past the cap won't get fixed.

appreciate it, directions 100% right. sort the tests + dedup and i'll take another look.

Lenochxd:
Understood, I'll work on this asap :)
For `&` and `/`, this is not an issue at all, the tool searches for the full string before trying to split.

https://github.com/Lenochxd/SoulSync/blob/bf2c12606f2b03671b2057d16e062bf86eb58696/core/repair_jobs/comma_artist_splitter.py?plain=1#L204-L211
https://github.com/Lenochxd/SoulSync/blob/bf2c12606f2b03671b2057d16e062bf86eb58696/core/repair_jobs/comma_artist_splitter.py?plain=1#L367-L379

Adding a parameter for this would still be a great addition to the tool.

Lenochxd:
I think the PR is ready now, a dry run option would be great but i'm too lazy to make it right now

Lenochxd:
lgtm

Nezreka:
dedup's sorted and the symbols toggles are nice, thanks for going back at it. few more, all cause this writes to people's files:

tests are still red, 3 of them:

```
test_split_artist_parts                       missing arg 'symbols'
test_fix_no_tracks_resolves_as_already_gone   NameError: 'f' is not defined
test_bulk_fix_comma_artist_split_end_to_end   'Camellia, Toby Fox' != 'Camellia; Toby Fox'
```

look like test bugs not code bugs, but give those a test and when green we should be good there.

that last one is real though. no file list means the fixer returns success + `already_gone` and bulk_fix counts it as fixed, so it reports 1 fixed on a file it never touched. current findings store artist_id and no paths, so on upgrade every existing one silently resolves having done nothing. skip them or don't count them.

dry run needs a rethink. true = findings only, false = auto apply everything with no findings at all. so unchecking "only reports issues without modifying files" retags the whole library, no findings, no record, weekly. either drop it and keep the findings flow, or rename to "apply automatically" defaulting off.

still want the band name asserts. i buy the full string lookup, your links made sense, but there's a toggle now and nothing proves it.

small one: rate limit pause is in the findings loop, api lookups are in the file loop. less bad than i said, memo makes it per unique artist not per file, but worth moving.

sort those and i'll take it, tag reading was the right call.

Lenochxd:
tests passed, please review

Nezreka:
merged, thanks for this. held up great against everything since july. no
conflicts, your tests pass, and the whole repair suite passes on the merged
result.

dry_run defaulting to true is the right call, and the whitelist for names like
Tyler, The Creator is exactly right. glad the tests use real files now too.

one small thing i fixed on dev instead of bouncing it back: a finding whose
files are all gone used to close itself when you hit fix (already_gone
success). the new empty-list return made that an error, so those findings got
stuck. restored the old behavior for that case. shout if it was deliberate.

sorry it sat so long.

PR #1103, opened 2026-07-30
Fix quality profile settings not persisting to quality_profiles table

Jean-LucPicard72:
## Summary
`_write_default_quality_profile_row()` only writes 6 of the 14 columns the Settings → Quality page can edit. Toggling `acoustid_required`, `downsample_enabled`, `deep_audio_verify`, `replace_lower_quality`, or any of the `lossy_copy_*` fields returns `{"success": true}` from `POST /api/quality-profile`, but the write-through to the `is_default=1` row in `quality_profiles` — which `get_quality_profile()` / `GET /api/quality-profile` actually reads — silently drops those columns. The UI/API keeps reporting the old value and the setting never takes effect.

Repro:
1. `POST /api/quality-profile` with `replace_lower_quality: true` (full profile object) → `{"success": true}`
2. `GET /api/quality-profile` → `replace_lower_quality` is still `false`
3. Confirmed directly in `quality_profiles` table: the row's `replace_lower_quality` column is unchanged.

## Fix
Adds the missing columns (`acoustid_required`, `downsample_enabled`, `deep_audio_verify`, `replace_lower_quality`, `lossy_copy_enabled`, `lossy_copy_codec`, `lossy_copy_bitrate`, `lossy_copy_delete_original`) to the `UPDATE` statement and parameter tuple, mirroring the same coercion style already used for the existing columns.

## Test plan
- [x] Manually verified: patched the same statement locally, `POST` a profile with `replace_lower_quality: true`, confirmed `GET /api/quality-profile` now returns `true` and the `quality_profiles` row reflects it.
- [ ] Would appreciate a maintainer sanity-check on the `lossy_copy_bitrate` type coercion (stored as `TEXT` in schema; profile payload can send it as either a string or number depending on client).


Nezreka:
Sorry I will be reviewing PR soon, trying to wrap up this react migration.

nick2000713:
@Jean-LucPicard72 thanks for digging into this.. the bug is real, i reproduced it exactly as you wrote it. and it is my own mess, i wrote that helper when i first implemented quality profiles.. the patch would hurt people though, so here is what i found before anyone merges it.

the eight columns are not missing by accident though. they have a different owner. `sync_default_quality_profile_from_config()` writes them onto the `is_default` row from the global config, after every settings save that touches `acoustid`, `lossy_copy`, `post_processing` or `import`. ladder comes from the profile payload, those eight come from the config side.

and `collectQualityProfileFromUI()` in settings.js posts only the six that are already written. the other eight are never in that body. so with your patch `profile.get("acoustid_required")` is `None` and lands in the row as `0`.

same db, same starting row, once without and once with the patch:

```
before        acoustid=1 downsample=1 deep_verify=1 replace_lower=1
              lossy=1 codec=opus bitrate=256 delete_original=1

dev today     unchanged
with patch    acoustid=0 downsample=0 deep_verify=0 replace_lower=0
              lossy=0 codec=mp3 bitrate=320 delete_original=0
```

that fires on any normal save on Settings → Quality. and `debouncedSaveQualityProfile()` hangs on four controls in the ladder editor, so just reordering your formats does it. presets too, `get_quality_preset()` only returns five keys.

so you set up lossy copy as opus 256 with delete original, come back to the quality page a week later, and it is all quietly off. stays off silently as well.. the pipeline reads `profile.get('deep_audio_verify', config.get(...))`, and after the wipe the key is not missing, it is `False`, so the config fallback never kicks in.

i also applied your patch and ran `tests/quality`. 390 passed. nothing in there looks at those eight.

your complaint stands though. `GET` hands back fourteen fields and `POST` quietly eats eight, that is a bad contract either way. so i fixed it in #1169 from the other side: the update is built from the ladder columns plus whichever bundle columns you actually put in the body. presence decides, not truthiness, so a field can go off and not only on. six field body leaves the rest alone, full `GET` payload posted back is a no op. your repro passes. coercion moved onto `_quality_profile_bundle_params` which create and update already use, so there is one place to add a column now instead of three. five tests came with it.

on your bitrate checkbox: `str(... or "320")` matches what `_quality_profile_bundle_params` does and the column is TEXT, so that part was fine either way. it is the `or` that hurts, a missing key becomes a written default.

nick2000713:
@Jean-LucPicard72 , the fix has been merged. You can try it out now...

PR #1121, opened 2026-08-04
Fix/pre lv2 bugfixes v2

### LINKED ISSUE #1109: Upgraded track land in library root folder not artist/album
Hey together, I have a problem: I just want to upgrade the quality of a album i have with the respective functionality. It downloads it correctly, but then it gets moved to to the root of my library (Transfer folder) and not sorted into the artist/album structure. For other initial downloads that worked without issues. Does it work for you or is it a bug?

nick2000713:
# Bugfixes that already affect `dev` — split out of the Library v2 PR

I said in #1062 that a bunch of the fixes I made along the way probably don't belong in that PR. This is me actually doing that. Every change in here fixes something that is broken on `dev` **today** and has nothing to do with Library v2 — no new tables, no new UI, no feature flag, nothing you have to opt into.

I cut it against current `dev`, so it can go in on its own whenever you're happy with it. Library v2 then gets that much smaller and, more importantly, honest: what's left in #1062 is actually Library v2.

**57 files, +2,324 / −383.** ~5,000 backend tests green plus the webui vitest suite.

---

## Why these were sitting in the wrong PR

I found nearly all of these by accident. I'd be testing something in the new library, hit behaviour that made no sense, dig in, and land on a bug that had been there the whole time and had nothing to do with what I was building. The async event-loop one below is the clearest example — I spent a day convinced I'd deadlocked my own code.

They ended up in the Library v2 branch purely because that's where I was working.

---

## The runtime one, which I think is the important one

`utils/async_helpers` runs one shared event loop on a daemon thread and every sync caller crosses into it through `run_coroutine_threadsafe`. On Python 3.14.6 that wakeup can get lost in a long-lived process — the loop sits in `select()` forever while the caller blocks on `future.result()` with no timeout. No error, no log line, no traceback. The subsystem just stops.

I couldn't reproduce it reliably enough to call it, so I fixed the shape instead of chasing the symptom:

- the loop is now **created inside the thread that runs it** (creating it in the caller and installing it in the thread is what leaves the selector unable to wake),
- jobs cross a `queue.Queue` that a pump task drains, instead of relying on that wakeup at all,
- the pump task is held by a strong reference and has a done-callback. It used to be neither, so if it ever died, the thread stayed `is_alive()` while nothing drained the queue — a permanent silent hang with an eventual GC warning as the only evidence,
- `run_async` takes an optional `timeout`. Default stays unbounded because legitimately long jobs go through here, but any caller that holds a lock while waiting on an external client now passes one.

Alongside it, `core/async_utils` gives provider I/O and download-control calls **separate** bounded pools. They shared the default executor before, so a slow metadata provider could starve download control.

---

## Prowlarr — this is the "Usenet finds nothing" report

Four separate things, all of which make a working setup look broken:

- **The 15s search timeout was a hard constant.** No setting could raise it. A search fans out to every enabled indexer and waits for the slowest, so it fired routinely. It's 75s now and honours `download_source.source_search_timeout` — which was already wired into HiFi/Qobuz/Deezer search and just never reached Prowlarr.
- **A failed search returned `[]`**, exactly like a genuine zero-hit search. So a source that failed *every single time* looked like "nothing matched". It raises `ProwlarrSearchError` now.
- **Category 3060 (Audio/Foreign) was excluded.** That's where most indexers file J-Pop/K-Pop/C-Pop/Bollywood. Those releases were structurally unfindable over Prowlarr while Soulseek — which has no category filter — kept finding them. So the gap read as "Usenet is broken for this artist". 3030 (Audiobook) and 3020 (Video) stay out; those really are different media.
- **`prowlarr.indexer_ids` is one shared setting for both protocols.** Fill it with torrent indexer IDs and the usenet plugin searches torrent-only indexers forever — zero results, no error, no way to tell from the UI. It's narrowed per protocol now, and an allowlist a protocol can't satisfy falls back to "all enabled indexers of that protocol" rather than returning nothing.

Plus: a query that comes back empty is retried with progressively relaxed variants before giving up.

---

## Two that can lose a file

- **`safe_move_file` treated "the destination exists" as proof that a failed move had actually committed** — and deleted the source. But on a replace the destination is the *existing library file*; it exists either way. Only a vanished source proves the rename went through, so that's what it checks now.
- **Quarantine wrote to the unresolved download path** while every other quarantine consumer (retry, cleanup, approve, list, clear) resolves it through `docker_resolve_path` first. On Docker with a Windows-style path configured, entries landed somewhere nothing could find them again. And the filename was `{second-resolution timestamp}_{name}` — two candidates for the same track quarantined in the same second is the *normal* shape of a multi-candidate retry walk, and `safe_move_file` overwrites its destination, so the loser and its sidecar just disappeared.

---

## Fixes #1109 — upgraded track lands in the library root

This one has a proper report attached, so in detail:

Artist-Enhance records the track's library path as `original_file_path` and writes the replacement back into that folder. That path is the **media server's** view of the library (`/music/…`) — which the SoulSync container usually doesn't have mounted at all. So `build_final_path_for_track` tried to `makedirs('/music')` and raised `PermissionError`. Post-processing aborted before it moved anything, the finished download was left sitting unsorted, and the retry loop hit the same crash on every pass. That's the exact traceback in the issue.

The recorded path now goes through the shared library-path resolver, and it only replaces in place when the folder is genuinely reachable. Otherwise it falls through to the normal path template, so the upgrade still lands in Artist/Album under Transfer instead of crashing.

There's a second half to it: the old-file retirement in the pipeline didn't resolve the path either, so `os.path.exists` said `False` and it cheerfully logged "Replaced in-place" while the superseded copy stayed on disk next to the new one.

---

## Metadata

- **`save_audio_file` used a fixed `.sstmp` temp name.** That's only safe if writers to one file are serialized, and they aren't — a bulk-fix thread and a scan's own auto-fix reach the same file at the same time. The loser fell into the generic except branch and dropped back to the non-atomic in-place save this function exists to replace.
- **The artist-image picker had no time budget.** One rate-limited provider sleeping tens of seconds inside its worker thread blocked the whole request past the frontend's client timeout, throwing away *every* source's result including the ones that had already come back. It's bounded now, and results are collected in configured provider order — iterating the completed set let a faster fallback source nondeterministically steal the preferred source's card.
- **Same-name artist disambiguation.** Providers carry fragment entries under the exact same name — Deezer lists five "Hiroyuki Sawano"s, and the first exact hit had 4 albums while the real one had 104. Among equal exact matches, catalog size wins now, then follower/popularity counts. No signals reported = the old first-exact-hit behaviour, unchanged.
- Version qualifiers are shared between title matching and audio verification, so `Title (X Remix)` and `Title - X Remix` can't normalize to different identities anymore.

---

## Auto-import

Exactly-once approval (an approved item could be picked up twice), truthful end states (`_process_matches` returned a bare bool, so a partial import reported as a clean success), manual-import containment to the allowed staging/download directories, and the approve endpoints now kick a scan so an approval isn't waiting on the next tick.

Auto-import predates the Library v2 branch entirely — these fixes were never Library v2's to carry.

---

## Web UI

- `api-client` ignored the standard `{error: {code, message}}` envelope and showed the generic HTTP message instead of the server's actual reason. So "Wishlist processing is already running." rendered as "Conflict".
- Dialogs take an `initialFocus`, and the close button has a title tooltip.

---

## On the relationship to #1062

This branch is a strict subset of `library-overhaul` — I verified that by test-merging it back in. It produces exactly two trivial conflicts in `core/imports/pipeline.py`, both "keep the Library v2 side", and after that the trees are identical apart from a handful of extra tests.

So merging this first costs #1062 nothing. It just means that when you review Library v2, you're reviewing Library v2.


Nezreka:
sorry for the delay here. a lot of genuinely good fixes in here. two things
will break stuff though, and i can reproduce both.

## blockers

**1. `core/matching/audio_verification.py` — the dash-qualifier rewrite eats real titles.**

the old regex required the *whole* qualifier to be a version word
(`vocal|instrumental|acoustic|live|remix|...`). `is_version_qualifier` returns
true if *any* token is a marker. so:

```
                              dev                    this pr
'Queen - Radio Ga Ga'    'queen radio ga ga'   ->   'queen'
'Billy Joel - Piano Man' 'billy joel piano man'->   'billy joel'
'Beyonce - Single Ladies''beyonce single ladies'->  'beyonce'
'Wings - Live and Let Die'                     ->   'wings'
'a-ha - Take On Me'                            ->   'aha'
```

radio/piano/single/live/take are all in `_VERSION_MARKER_TOKENS`. `'Song - Live'`
and `'Song - Remix'` still work, so the intent survives — it just over-matches
badly. this feeds acoustid/title match decisions so it'll mis-verify real tracks.

needs the qualifier to be *entirely* version words, not contain one.

**2. `core/imports/file_ops.py:118` — cross-device imports land 0600.**

`tempfile.mkstemp` creates at 0600 by design. the old `open(tmp, "wb")` got
0644 via umask, and `os.replace` preserves the source mode:

```
umask-respecting open(): 0o644
tempfile.mkstemp()     : 0o600
after os.replace       : 0o600
```

same-filesystem moves use `os.replace` directly so they're fine. cross-device is
exactly the docker downloads-volume → library-volume case, and plex/jellyfin
usually run as a different user. they won't be able to read anything imported.

`shutil.copymode(src, tmp)` before the replace, or chmod explicitly.

## should fix

**3. `core/async_utils.py:38` — busy poll.** `while not worker.done(): await
asyncio.sleep(0.05)` is a 50ms latency floor and 20 wakeups/sec per call.
`await asyncio.wrap_future(worker)` does the same job with none of that. the
`worker.cancel()` on line 42 is also a no-op — concurrent futures can't cancel
once started.

**4. `utils/async_helpers.py` `_pump_jobs` — 100hz poll for a bug we don't hit.**
the queue+pump machinery works around `run_coroutine_threadsafe` on python
3.14.6. docker and ci are 3.11, dev is 3.12.3. gate it on `sys.version_info` or
drop it. (the node 26 localStorage shim in vitest.setup is fine — local node
really is 26.7.0.)

**5. `core/download_plugins/torrent.py` `prowlarr_search_with_variants`** —
computes `usable = [r for r in results if r.protocol == protocol]` then
`return results`. so the protocol filter only decides whether to stop retrying,
it doesn't filter output. probably deliberate since `_project_results` filters
later, but it reads like a bug. one comment would fix it.

**6. `core/imports/routes.py` `_validate_import_file`** — imports are now
restricted to staging + download_path. good hardening, but anyone importing from
a third folder starts getting "file is outside the allowed import folders". is
that a workflow anyone uses?

**7. `_process_matches` now appends an error per unmatched file**, and
'completed' requires `not errors`. so an album with any extra audio file lands
'partial' forever — and `_is_already_processed` now includes 'partial', so it
won't retry. intentional?

## nits

- `query_variants.strip_bracket_groups` docstring says an unbalanced opener is
  "left in place"; the code replaces it with a space. unbalanced *closers* do
  survive. harmless, just doesn't match the docs.
- `_drop_noise_tokens` drops "with" — real word in plenty of titles. can only
  widen an AND search so it won't lose hits, but it's the one aggressive entry.
- `prowlarr_search_with_variants` normalizes `protocol` *after* passing it to
  `indexer_ids_for_protocol`. harmless (that method normalizes internally) but
  confusing to read.
- `auto_import_worker` imports `import_rejection_reason` inside the per-track
  loop.
- `gather_artist_image_candidates` — `pool.shutdown(wait=False)` leaves
  stragglers running, and the worker cap went from `min(len, 6)` to `max(len, 1)`.

## the good stuff

- `safe_move_file`'s `destination_preexisted` logic is a real bug fix — the old
  code treated "destination exists with size > 0" as proof the move succeeded
  and deleted the source, when it could just be the old library file.
- quarantine: `docker_resolve_path` + the collision suffix. both real.
- `save_audio_file`'s unique temp name — fixes concurrent writers falling back
  to the non-atomic path.
- `_reachable_original_dir` for #1109 — not creating `/music` is right.
- prowlarr: splitting search (75s) from api (15s), `ProwlarrSearchError`,
  `indexer_ids_for_protocol`, adding category 3060. all well reasoned.
- `_pick_best_artist_match` catalog weighting, and ties still return the first
  hit so old behaviour holds when a source reports nothing.
- `approve_item` on one connection with an atomic rowcount check.
- the whole `get_event_loop().run_in_executor(None, ...)` cleanup across the
  adapters — that was deprecated and needed doing.
- artist image: iterating `future_by_source` instead of the `done` set so a fast
  fallback can't steal the preferred source's slot.

full python suite is running on the merged branch, i'll post if anything falls
over.

nick2000713:
@Nezreka No worries about the delay. Thanks for reproducing both of those, saved me a lot
of guessing.

All fixed and pushed. Two of them didn't go the way you expected, so here's what
I found.

## 1. the dash qualifier

You're right, and it's worse than cosmetic. `normalize()` feeds the AcoustID
title decision, so `'Queen - Radio Ga Ga'` and `'Queen - Live'` both collapsed
to `queen` and compared at 1.00. That's a wrong recording passing verification.

Where I went a different way is the fix. "Entirely version words" drops the
cases the rewrite was for:

```
'King Of My Castle - Don Diablo Edit'    don, diablo    not markers
'Monster - Robin Schulz Remix'           robin, schulz     "
'void - super slowed'                    super             "
```

Those are the reported quarantines the symmetry tests are built from, so that
rule just swaps your false positives for the old false negatives.

What splits the two groups is where the marker sits. A version tag ends on its
marker, a real title only starts with one:

```
'Radio Ga Ga'         last token: ga         kept
'Piano Man'           man                    kept
'Single Ladies'       ladies                 kept
'Take On Me'          me                     kept
'Don Diablo Edit'     edit                   stripped
'2011 Remaster'       remaster               stripped
'Original Mix'        mix                    stripped
```

So it's `is_trailing_version_qualifier` now, and only the dash path uses it. The
bracketed path keeps "contains a marker", which is fine there because brackets
aren't ambiguous the way a dash is. All five of your examples are pinned, plus
the predicate gets its own tests in `tests/text/test_version_qualifier.py`.

Two things I left alone. A real title ending on a bare marker word ("... Mix")
still gets stripped, the old regex did that too. And a CJK tail has no ASCII
tokens at all, so it's never treated as a version: `'YouSeeBIGGIRL/T:T - 澤野弘之'`
keeps the whole string, `'進撃の巨人 - Live'` still strips. Same as before, but I
checked rather than assumed.

## 2. the 0600 one

I can't reproduce it. `_atomic_cross_device_move` calls
`shutil.copystat(src, tmp)` before the `os.replace`, and has done since
`9d091207f`, which is in dev and untouched by this PR. So the mode comes from
the source:

```
0644 -> 0644
0664 -> 0664
0600 -> 0600
```

New parametrised test, passes on the branch exactly as you reviewed it. Easy one
to miss, the copystat line is context, not part of the hunk.

There is a real hole right next to it though. copystat does the mode first, then
timestamps and xattrs, so if the utime step raises, the chmod goes with it and
the file publishes at 0600. Same symptom, different route. `copymode` now runs
first on its own, with a test that pins the mode while copystat raises.

## the rest

**3. busy poll.** Done, it's `wrap_future` now, and `worker.cancel()` went with
it. 50 sequential calls: 0.50s to under 0.01s.

**4. `_pump_jobs`.** Dropped it instead of gating it. The real bug was the loop
being created in the calling thread, which this PR already fixed in `_run_loop`,
so I retested the plain path on 3.14.6: 3200 coroutines from 8 threads and a
submission after 45s of idle `select()`, zero lost wakeups. About 90 lines gone,
and `run_async(..., timeout=)` now actually cancels the coroutine instead of
just giving up on it. Two tests cover the invariant rather than the mechanism.

**5. `usable`.** Returns `usable` now. Callers all re-filter by protocol anyway,
so nothing downstream changes. `protocol` normalisation moved above
`indexer_ids_for_protocol` too, that was your third nit.

**6. import roots.** I don't think a third folder is reachable: the only
producer of `full_path` is `staging_files`, and there's no path-entry UI, so the
client can't name a path it wasn't handed. What does break is a staging tree of
symlinks pointing elsewhere, which is deliberate (this endpoint moves the file)
and pinned by a test, but the message was a dead end, so a rejection now logs
the roots. Happy to make them configurable if you'd rather, one line in
`_get_allowed_import_roots`.

**7. unmatched files.** Fixed, and worse than you thought: `unmatched_files`
also carries the quality-dedup losers, so the lower bitrate copy of a track that
*did* import counted as a failure and parked a healthy album at 'partial'
forever. Only a failed track blocks 'completed' now. 'partial' stays terminal in
`_is_already_processed` on purpose, the landed tracks are already out of the
folder so a re-run would read the remnant as a new album. Comment in the code.

## nits

- `strip_bracket_groups` docstring now matches the code.
- `_drop_noise_tokens` "with": kept with a comment. Dropping a token only widens
  an AND search and the exact query already ran as variant 1, so it can't lose a
  hit. Say the word and I'll drop it anyway.
- `import_rejection_reason` hoisted to module scope, no cycle.
- `gather_artist_image_candidates`: both deliberate, both commented now.
  `shutdown(wait=False)` because waiting just moves the stall into the request
  (non-daemon threads, so they're joined at exit). One worker per source because
  the budget is wall clock for the whole fan-out: under a cap of 6, source 7
  doesn't get a slow start, it gets no start, and a provider quietly vanishes
  from the picker depending on its place in the chain.

`tests/imports` 745 green plus text/matching/utils/prowlarr/async, and the
cross-script/CJK set (171) since that path is downstream of the qualifier
change. Curious what the full suite says on your side.

PR #1122, opened 2026-08-05
Use configured search timeout instead of hardcoded 15 seconds for Prowlarr searches

techlover112:
The search timeout setting that was recently added didn't apply to Prowlarr. This PR sets it to use that config, with a fallback of 60 seconds (Prowlarr usually returns results around the 45 second mark in my experience). I haven't been able to test this, but it should be fairly stable, as I just copied the config block from deezer_download_client.py.


techlover112:
Want to merge into dev instead of main

Issue #1056, opened 2026-07-20
Enhancement: Settings option to edit individual source timeout thresholds


techlover112:
I would really appreciate this. Right now I have to edit the internal web source code in order to tell SoulSync not to time out after just 15 seconds waiting on a search query.

Nezreka:
I have this live on dev now, will push main soon

PR #1123, opened 2026-08-05
Use configured source search timeout instead of hardcoded 15 seconds

techlover112:
Uses the new source search timeout config instead of the hardcoded 15 seconds. I haven't been able to test this, but it should be fairly stable, as I just copied the config block from deezer_download_client.py. The fallback is 60 seconds, as Prowlarr usually returns results around the 45 second mark in my experience.

### DIFF
--- a/core/prowlarr_client.py
+++ b/core/prowlarr_client.py
@@ -100,8 +100,6 @@ class ProwlarrSearchResult:
 class ProwlarrClient:
     """Thin sync-backed async wrapper around the Prowlarr v1 API."""
 
-    DEFAULT_TIMEOUT = 15
-
     def __init__(self) -> None:
         self._load_config()
 
@@ -243,7 +241,7 @@ class ProwlarrClient:
                 url,
                 headers={'X-Api-Key': self._api_key, 'Accept': 'application/json'},
                 params=params,
-                timeout=self.DEFAULT_TIMEOUT,
+                timeout=getattr(self._config, 'get_source_search_timeout', lambda: None)() or 60,
             )
             if not resp.ok:
                 logger.warning("Prowlarr %s returned HTTP %s", path, resp.status_code)


Nezreka:
good catch on the hardcoded timeout, but this crashes as written.

`self._config` doesn't exist on ProwlarrClient. `_load_config` only sets
`self._url` and `self._api_key`, it uses the module-level `config_manager`. so
every prowlarr search throws `AttributeError: 'ProwlarrClient' object has no
attribute '_config'`.

the getattr doesn't save you there — it guards a missing attribute *on* the
object, but python evaluates `self._config` first and blows up before getattr
ever runs.

also the fallback should be 15, not 60. the #1056 contract is that an unset
override changes nothing out of the box, so each call site keeps its own old
default. prowlarr's was `DEFAULT_TIMEOUT = 15`. 60 quadruples it for everyone
who hasn't set the override.

so:

```python
timeout=config_manager.get_source_search_timeout() or 15,
```

same as qobuz and hifi. `config_manager` is already imported in that file.

last thing — `tests/test_source_search_timeout.py` has a seam pin for every
other call site (hifi, qobuz, deezer, stream). add one for prowlarr so this
doesn't drift.

nick2000713:
This PR  and mine #1121 fix the same thing...the hardcoded 15s in
`core/prowlarr_client.py` and even read the same setting
(`config_manager.get_source_search_timeout()`). Not saying yours is wrong, I
just don't want either of us doing the work twice, and whoever merges second is
going to hit a conflict in that file.

Your 45s observation matches what I measured, which is why I ended up at 75
rather than 60.

One difference worth weighing: this puts the search budget on `_api_get`, and
every Prowlarr call goes through there.. including `system/status` and the
indexer list. Those answer immediately, so a Prowlarr that's down or
misconfigured would make the connection test in Settings sit for the full 60s
instead of failing in 15.

#1121 splits the two instead:

- `DEFAULT_TIMEOUT = 15` stays for metadata/control calls (status, indexers)
- `DEFAULT_SEARCH_TIMEOUT = 75` for searches, which fan out to every enabled
  indexer and wait for the slowest one
- precedence is caller argument > `download_source.source_search_timeout` (the
  same user setting you're reading) > default, in
  `ProwlarrClient.resolve_search_timeout()`
- the plugins' own `timeout=` argument actually reaches Prowlarr now — it was
  being accepted and then dropped
- a search that times out raises `ProwlarrSearchError` instead of returning
  `[]`, so a failing indexer stops looking identical to "no results"

If #1121 goes in first this is fully covered by it and can close; if this lands
first I'll rebase #1121 on top of it. Either way is fine by me — @Nezreka's
call.

techlover112:
@Nezreka @nick2000713 Thanks for your help. I'm going to close this so #1121 will merge easily.

PR #1125, opened 2026-08-06
feat: add genre enrichment repair job

mandos21:
## Summary

  Adds a disabled-by-default `genre_enrichment` repair job that identifies missing or incomplete artist and album genres and proposes
  safe, reviewable additions.

  The job uses existing library metadata and cached provider payloads first, supports optional live enrichment by existing provider
  IDs, translates provider genre names against the configured whitelist, and never removes existing canonical genres.

## Screenshots
image
image
image
image
image
image

  ## What changed

  ### New genre enrichment service

  Added `core/metadata/genre_enrichment.py`, providing reusable functionality for:

  - Reading the configured genre whitelist
  - Hashing whitelist versions for translation-cache isolation
  - Parsing genre values from database fields and provider payloads
  - Collecting local and metadata-cache candidates
  - Extracting provider-specific genres
  - Translating provider genres to whitelist values
  - Ranking and deduplicating candidates
  - Applying per-entity genre limits
  - Preserving provenance, scores, ambiguity, and rejection details

  Supported provider mappings include:

  - Spotify artist and album `genres`
  - iTunes `primaryGenreName` and equivalent genre fields
  - Deezer nested genre payloads and genre ID mappings
  - Discogs `genres` and `styles`
  - Last.fm stored tags
  - AudioDB `strGenre`

  MusicBrainz is intentionally not treated as a genre source until its existing metadata payload shape is explicitly mapped.

  ### New repair job

  Added and registered `genre_enrichment` with:

  - Display name: `Genre Enrichment`
  - Default state: disabled
  - Default interval: 168 hours
  - Auto-fix: disabled
  - Default maximum: 5 genres per artist or album

  Available settings:

  - `max_genres`: integer from 1 through 20
  - `include_artists`
  - `include_albums`
  - `allow_live_calls`

  Invalid settings are normalized to safe defaults server-side.

  The job skips cleanly when the genre whitelist is disabled or empty.

  ### Non-destructive genre behavior

  The job:

  - Preserves all existing canonical genres
  - Adds only translated whitelist genres
  - Avoids duplicates using shared normalized comparison
  - Stops adding when the configured maximum is reached
  - Leaves entities with existing genre counts above the cap unchanged
  - Reports genres omitted because of the cap
  - Does not rewrite audio files

  The fix handler re-reads the current database value before applying changes, preserving genres added after the scan and avoiding
  stale overwrites.

  ### Candidate-source and cache behavior

  Candidates are collected in this order:

  1. Existing canonical genres
  2. Stored Discogs and Last.fm fields
  3. Cached provider entity payloads
  4. Optional live provider detail lookups

  Live lookups only use existing provider IDs. The job does not perform speculative name searches.

  Cached entities are checked before making live calls. Provider failures are isolated to the affected entity/source so one provider
  failure does not abort the complete library scan.

  ### Persistent translation cache

  Added the `genre_translation_cache` table to the normal database initialization path.

  The cache stores:

  - Accepted translations
  - Ambiguous matches
  - Rejected/no-match values
  - Whitelist hash
  - Normalized source genre
  - Matched genre
  - Score
  - Margin
  - Top candidate list
  - Access timestamps and counts

  The whitelist hash ensures translations are not reused after the configured whitelist changes.

  A per-run in-memory cache also prevents duplicate translation work during a single scan.

  ### Conservative genre matching

  Extended shared genre normalization to support:

  - Case-insensitive matching
  - Unicode normalization
  - Hyphen and dash variants
  - Ampersand variants
  - Common punctuation
  - Repeated whitespace
  - Safe aliases such as:
    - `hip-hop` → `Hip Hop`
    - `HipHop` → `Hip Hop`
    - `RnB` → `R&B`
    - `R&B` → `R&B`

  For values that do not match exactly, the service uses `SequenceMatcher` combined with token overlap.

  Results are classified as:

  - Accepted for high-confidence matches
  - Ambiguous when a plausible match exists but confidence or margin is insufficient
  - Rejected for low-confidence values

  Ambiguous values are surfaced for review rather than silently applied.

  ### Candidate ranking

  Accepted genres are ranked using:

  1. Number of independent provider sources
  2. Provider confidence
  3. Translation confidence
  4. Original provider order
  5. Stable normalized alphabetical ordering

  Provider confidence weights include:

  - Discogs: `0.90`
  - Spotify: `0.85`
  - Deezer: `0.80`
  - Last.fm: `0.75`
  - iTunes: `0.70`
  - AudioDB: `0.65`
  - Other/embedded sources: `0.60`

  Genres supported by multiple providers receive higher priority.

  ### Findings and repair actions

  The job creates one finding per artist or album when it has:

  - At least one applicable genre addition, or
  - An ambiguous candidate requiring review

  Finding details include:

  - Current genres
  - Proposed genres
  - Added genres
  - Ambiguous candidates and scores
  - Rejected values
  - Genres omitted due to the cap
  - Provider provenance
  - Cache and live-call statistics
  - Maximum genre setting

  Added the `genre_enrichment` fix handler, which:

  - Validates the finding
  - Re-reads the current entity
  - Merges additions without deleting existing genres
  - Deduplicates values
  - Fails safely if the entity no longer exists
  - Resolves the finding only after a successful database update

  ### Maintenance UI

  Updated the existing repair-job UI to support:

  - Genre Enrichment job display
  - Genre enrichment settings
  - `Apply Genres` action
  - `Genres Applied` resolution label
  - Current and proposed genres
  - Added, ambiguous, rejected, and cap-omitted genres
  - Provider source information
  - Cache and external-call counts
  - Fuzzy match scores

  No new API endpoints were required.

  ## Safety and rollout

  - Disabled by default
  - Live API calls disabled by default
  - Existing genres are never removed
  - No audio files are modified
  - The whitelist must be enabled and non-empty
  - Provider calls only use existing source IDs
  - Findings must be reviewed before applying changes

  The existing `genre_cleanup` job remains unchanged and continues to provide removal-only whitelist cleanup.

  ## Tests

  Added focused tests covering:

  - Case-insensitive exact matching
  - Hyphen and punctuation normalization
  - `HipHop`, `hip-hop`, `R&B`, and `RnB` aliases
  - Ambiguous and rejected fuzzy matches
  - Stable ranking
  - Provider support weighting
  - Maximum genre behavior
  - Preservation of existing genres above the cap
  - Deezer genre ID mapping
  - Cached provider ID lookup

### DIFF
--- a/core/genre_filter.py
+++ b/core/genre_filter.py
@@ -4,6 +4,8 @@ When strict mode is enabled, only genres on the whitelist pass through
 during enrichment. When disabled (default), all genres pass unchanged.
 """
 
+import re
+import unicodedata
 from utils.logging_config import get_logger
 
 logger = get_logger("genre_filter")
@@ -95,8 +97,10 @@ DEFAULT_GENRES = [
     "Lounge", "Psychedelic", "Progressive",
 ]
 
-# Normalized lookup set — built once, used for O(1) matching
-_DEFAULT_LOOKUP = {g.lower() for g in DEFAULT_GENRES}
+_GENRE_ALIASES = {
+    'hiphop': 'hip hop', 'hip hop': 'hip hop', 'rnb': 'r and b',
+    'r b': 'r and b', 'r and b': 'r and b', 'rhythm and blues': 'r and b',
+}
 
 
 def _normalize_for_match(genre: str) -> str:
@@ -104,10 +108,20 @@ def _normalize_for_match(genre: str) -> str:
 
     Handles common variations: 'R&B' vs 'RnB', 'Hip-Hop' vs 'Hip Hop'.
     """
-    g = genre.lower().strip()
-    g = g.replace('-', ' ').replace('&', ' and ')
-    # Collapse multiple spaces
-    return ' '.join(g.split())
+    if genre is None:
+        return ''
+    g = unicodedata.normalize('NFKC', str(genre)).lower().strip()
+    g = g.replace('\u2010', '-').replace('\u2011', '-').replace('\u2012', '-')
+    g = g.replace('\u2013', '-').replace('\u2014', '-').replace('\u2212', '-')
+    g = g.replace('&', ' and ')
+    g = re.sub(r'[-_/.,;:|]+', ' ', g)
+    g = re.sub(r"[’'\"`()\[\]{}!?]+", '', g)
+    g = ' '.join(g.split())
+    return _GENRE_ALIASES.get(g, g)
+
+
+# Normalized lookup set — built once, used for O(1) matching
+_DEFAULT_LOOKUP = {_normalize_for_match(g) for g in DEFAULT_GENRES}
 
 
 def _build_lookup(genre_list):
--- a/core/metadata/cache.py
+++ b/core/metadata/cache.py
@@ -138,6 +138,47 @@ class MetadataCache:
             logger.debug(f"Cache lookup error ({source}/{entity_type}/{entity_id}): {e}")
             return None
 
+    def get_genre_translation(self, whitelist_hash: str, normalized_source_genre: str) -> Optional[dict]:
+        """Return and touch a persisted genre translation outcome."""
+        try:
+            db = self._get_db(); conn = db._get_connection()
+            try:
+                cur = conn.cursor()
+                cur.execute("SELECT * FROM genre_translation_cache WHERE whitelist_hash=? AND normalized_source_genre=?",
+                            (whitelist_hash, normalized_source_genre))
+                row = cur.fetchone()
+                if not row:
+                    return None
+                cur.execute("UPDATE genre_translation_cache SET last_accessed_at=CURRENT_TIMESTAMP, access_count=access_count+1 WHERE id=?", (row['id'],))
+                conn.commit()
+                return dict(row)
+            finally:
+                conn.close()
+        except Exception as e:
+            logger.debug("Genre translation cache lookup failed: %s", e)
+            return None
+
+    def store_genre_translation(self, whitelist_hash: str, source_genre: str,
+                                normalized_source_genre: str, outcome: dict) -> None:
+        try:
+            db = self._get_db(); conn = db._get_connection()
+            try:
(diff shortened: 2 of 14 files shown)


Nezreka:
nice work, backend looks good. schema handling is right too.

heads up i moved the base from main to dev, so your frontend is stranded. not
your fault.

**blocker:** tools page is react on dev now (#1133). your enrichment.js edits
are on a dead page. so no apply button at all, and the detail panel comes up
basically empty.

same stuff you wrote, just in react:

`webui/src/routes/tools/-tools.core.ts` — three one-liners next to the
genre_cleanup ones:
- `FINDING_TYPE_LABELS` → `genre_enrichment: 'Genre Enrichment'`
- `FINDING_FIXABLE_TYPES` → `genre_enrichment: 'Apply Genres'`
- `FINDING_ACTION_LABELS` → `genres_applied: 'Genres Applied'`

`webui/src/routes/tools/-ui/finding-detail.tsx` — a `case 'genre_enrichment'`.
copy the genre_cleanup case above it. ~25 lines.

**also:**

1. applying an ambiguous-only finding says it worked but does nothing.
   added_genres is empty, loop does nothing, still returns genres_applied.
2. scan does `SELECT *` + fetchall on artists and albums. whole library in
   memory. iterate the cursor instead.
3. `_normalize_for_match` is shared with the live genre filter, so this changes
   whitelist behavior for everyone on strict mode. probably better, but can you
   add a few tests pinning the existing cases.

nits: `lastfm_id` column doesn't exist (only lastfm_tags) so that lookup is
always none. `collect_local_candidates` takes entity_type and never uses it.
branch has 6 merge commits from main, squash before merge.

merged it onto dev locally, genre tests all pass.

get the react side + the ambiguous apply sorted and this is good to go.

mandos21:
Made the requested changes, going to do some more manual testing tomorrow.

mandos21:
The spacing on the settings dropdown was messed up so I fixed that as well. Between screenshots I also put it in the correct category.
Before:
image
After:
image

mandos21:
@Nezreka It's ready for another look. Let me know what you think and if there's anything else it feels like it should have.

PR #1145, opened 2026-08-14
fix(matching): recognize a title stated twice as one title

nstrelow:
YouTube Music titles sometimes state the same title twice: a localized-script title plus its Latin transliteration/translation (`花に亡霊 - Ghost in a Flower`), or a decorated title restated plainly (`COLORS (Album Mix) - Colors (Ailu Mix)`). Providers only index the second half, so matching the raw title finds nothing.

**Fix:** `restated_title()` returns the Latin half only when it's demonstrably a restatement:
- **different scripts** — one half is Latin, the other is non-Latin (or has no Latin letter at all — an unrecognized script never gets returned as "the Latin half"). Excludes generic credits like "Live at Budokan" or "Official Music Video", which are real subtitles, not restatements.
- **same core** — both halves reduce to the same string once decoration (`(Long Ver.)`, `-Naruto Mix-`) is stripped.

Used as an *additional* match candidate in `canonical_source_track`, only when no artist-prefix match was found — never overrides the stronger existing signal.

**Impact:** on my 22-mirror library, rescued 38 of 181 previously-unmatched tracks.

**Tests:** `tests/test_source_title.py`, 64 cases (positive + negative, including script edge cases). Full suite: no new failures vs pristine `dev`.

### DIFF
--- a/core/text/source_title.py
+++ b/core/text/source_title.py
@@ -53,6 +53,7 @@ _NON_LATIN = re.compile(
     "֐-׿"              # hebrew
     "؀-ۿ"              # arabic
     "฀-๿"              # thai
+    "ऀ-ॿ"              # devanagari
     "]"
 )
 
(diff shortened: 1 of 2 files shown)


Nezreka:
This is great merging it now

PR #1147, opened 2026-08-14
feat(wishlist): optionally search unverified Wing It guesses

nstrelow:
When discovery finds no catalogue match, it stores a "Wing It" stub and skips the wishlist by design — a stub is an unverified guess, and the Wing It Pool is the surface for resolving them by hand. On a large mirrored library that pool can grow past what's practical to work through by hand (224 stubs / 185 unique tracks on my 22-mirror library) — and most ARE findable: 77 of 185 match at ≥0.7 confidence with the existing matching engine.

**Fix:** new `wishlist.wing_it_guesses` setting, default **off**. When on, a stub is wishlisted like any other track — but only when the source gave a real artist *and* title; a nameless stub (`"Unknown Artist"`) stays excluded either way. All the call sites that skip stubs (`sync_service`, `wishlist_readd`, `wishlist_failed`, `wishlist/processing.py`) now share one predicate (`core/discovery/wing_it.py::should_wishlist_stub`) so they can't drift apart.

Also removes two early-returns that made the setting have no effect for the flows most likely to touch wing-it stubs: `sync_service`'s `_skip_unmatched_wishlist` (driven by Wing It Sync) and `wishlist_failed.py`'s `batch.get('wing_it')` (Wing It download batches) both used to bypass the per-track gate entirely. `skip_wishlist_add` (organize-by-playlist) is a separate reason and keeps its own blanket skip — only the wing-it-specific bypass is removed.

Retry backoff: `record_failed_attempt` used to skip stamping any `wing_it_`-prefixed id. With this setting on, a stub genuinely can be on the wishlist, and an unstamped row keeps `retry_count` at 0 forever — which means `retry_backoff` never escalates it and every cycle burns a fresh search on a track that's failed for months, the exact waste that module exists to stop. Stamping is now unconditional; a stub that isn't on the wishlist simply updates no row.

## Tests

Full suite (`--ignore=tests/js`): **43 failed, 13645 passed** vs pristine `dev`'s **43 failed, 13626 passed** — failure sets diffed and identical, all pre-existing `test_vanilla_globals_resolve`. Targeted `tests/discovery` + `tests/test_wishlist_retry_backoff.py` + `tests/test_no_app_code_in_config_dir.py`: 379 passed.

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -793,6 +793,19 @@ class ConfigManager:
                 # Turn off to canonicalize casing to the metadata source.
                 "reorganize_preserve_casing": True,
             },
+            "wishlist": {
+                # When discovery finds no catalogue match for a track it stores a
+                # Wing It stub built from the source's own artist/title, and the
+                # sync deliberately keeps those OFF the wishlist — a stub is an
+                # unverified guess, and the Wing It Pool is where you resolve
+                # guesses by hand. On a large mirrored library that pool can hold
+                # hundreds of tracks, which is more than anyone will work through.
+                # Turn this on to let guesses be searched like any other wishlist
+                # track. Only stubs with a real artist AND title qualify; ones the
+                # source left nameless stay out either way. OFF by default —
+                # behavior is byte-for-byte today's.
+                "wing_it_guesses": False,
+            },
             "scripts": {
                 "path": "./scripts",
                 "timeout": 60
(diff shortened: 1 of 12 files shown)


Nezreka:
Hey sorry about this conflict. This morning I moved settings.py out of the config folder and into core to solve some latent issues.

nstrelow:
No worries I'll adapt sometime tomorrow :)

Also feel free to add a CONTRIBUTING.md to your repo and detail how you'd like to receive contributions and with what rules.

I am very shamelessly using Claude Code here and I do not read every code it generates, but I do some extra runs to review it (I also run coderabbit on my fork, before making PRs to this)

nstrelow:
@Nezreka this should be good now :)

PR #1154, opened 2026-08-16
feat(youtube): Improve quality matching and support YT Premium formats

lucasssvaz:
## Summary

- **YouTube search was a pile of video titles.** Hits had no reliable artist/album, so matching was weaker than Tidal/Qobuz. Search now uses the YouTube Music catalog (structured artist/title/album). A pasted Netscape `cookies.txt` is enough in Docker; the browser dropdown does nothing there.

- **Re-encode is now optional.** Default is unchanged: convert to MP3 320 after download. Turn it off in Settings to keep the original Opus/AAC (FFmpeg remux / stream copy). Leftover DASH `.webm` next to the audio is discarded so import does not see the container.

- **SoulSync pretended YouTube was MP3 320.** YouTube does not serve MP3. Streams are Opus/AAC. Missing formats now claim typical Opus 160. Cookies are not Premium — itag 774 (Opus ~256) and 141 (AAC 256) are stamped only when the player actually returns them. With re-encode on, ranking uses the converted file on disk, not a fake YouTube MP3.

- **Best-quality search mixed sources into one matcher.** A Soulseek peer first forced YouTube/Tidal through the P2P path; a closer YouTube hit could drop a Tidal FLAC via the itag confidence band. Each source family now uses its own checker. YouTube is probed and ranked on its own rows; other sources stay in the pool. A rejected YouTube profile no longer wipes everyone else.

- **YouTube lost to Soulseek even when itag 774 was the top target.** Best-quality walks used Soulseek’s old `quality_score` (FLAC 1.0 vs Opus 0.3). Ranking now follows the profile list: Opus ≥192 beats later FLAC/MP3. Hybrid order only breaks ties on the same rung.

- **Premium itags showed up in the probe, then the download 403’d.** Ranking used `skip_download` URLs that YouTube then refused. Download always extracts fresh. A 403 retry used to strip cookies, which made 774/141 disappear; retries now keep cookies, try `web_music` only, and skip dead itag URLs.

- **Completed YouTube rows showed no quality (or fake MP3-320).** Mutagen cannot read Opus bitrate, so the chip was blank. The label now uses the itag probe bitrate when the file is still Opus/AAC. A transcoded MP3 still shows as MP3, not the Opus claim. Already-completed rows keep their old chips; only new imports get the new labels.

- **Library Enhanced view showed a dash for Opus bitrate.** Ogg Opus has no header bitrate, so import stored `0` while completed-download chips already estimated from size/duration. The album table now fills that average on import, scan, and artist-detail load, and prefixes Opus/AAC/WMA with `~` (VBR average). MP3 CBR stays an exact figure.

## Testing

I'm running a custom built image in my NAS for the last few days, Everything seems to be working fine. I am able to download Premium Youtube Music files at the original high quality. All tests also pass locally.

image

image

### DIFF
--- a/core/download_orchestrator.py
+++ b/core/download_orchestrator.py
@@ -381,11 +381,15 @@ class DownloadOrchestrator:
             logger.warning(f"No results found for: {query}")
             return None
 
-        # 2. Filter and validate results
+        # 2. Filter and validate results. Score each source family with its
+        # own checker — a Soulseek peer first must not force Tidal/YouTube
+        # through the P2P path (and vice versa).
         _streaming_sources = ('youtube', 'tidal', 'qobuz', 'hifi', 'deezer_dl', 'lidarr', 'soundcloud', 'amazon')
-        is_streaming = tracks[0].username in _streaming_sources if tracks else False
+        streaming = [t for t in tracks if getattr(t, 'username', None) in _streaming_sources]
+        p2p = [t for t in tracks if getattr(t, 'username', None) not in _streaming_sources]
 
-        if is_streaming and expected_track:
+        stream_kept = []
+        if streaming and expected_track:
             # Score streaming results against expected track metadata
             from core.matching_engine import MusicMatchingEngine
             me = MusicMatchingEngine()
@@ -400,7 +404,7 @@ class DownloadOrchestrator:
             expected_is_version = any(kw in expected_title_lower for kw in _version_kw)
 
             scored = []
-            for r in tracks:
+            for r in streaming:
                 confidence, _ = me.score_track_match(
                     source_title=expected_title,
                     source_artists=expected_artists,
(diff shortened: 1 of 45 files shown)


Nezreka:
hey, this is really solid work. the honest opus/aac labeling, the itag table, the per-source validation split. all good stuff and stuff i want in. few things before i can merge:

**the blocker:** you've got `from config.settings import config_manager` in 4 files (youtube_client, youtube_cookies, candidates, plus a test), and 3 test files monkeypatch that same path as a string. that module got moved to `core.settings` a while back because it lived in the folder users mount. i'm guessing you still have the old file locally, which is why it works for you. on a clean checkout the import fails, the try/excepts swallow it, and your main features silently turn themselves off. transcode toggle does nothing, catalog search is always anonymous. 12 of your own tests fail on a clean tree (they're good tests btw, they caught it). i verified renaming to `core.settings` across those 7 files takes everything green. quick fix.

**two calls i want your take on:**

1. catalog-first search has no yt-dlp fallback when the catalog returns hits but the matcher rejects them all. an obscure remix that's on raw youtube but not in the music catalog now gets zero candidates where it used to be found. can we fall through to ytsearch when the catalog batch strikes out?

2. the old cookie-strip retry is gone. expired or bot-flagged cookies now poison all three attempts. the old code dropped cookies on a retry for exactly that reason. was removing it intentional?

sorry about that merge conflict.

lucasssvaz:
Thanks for the review. Merged the dev branch and solved conflicts. About the changes:

`catalog → ytsearch`: Agreed. Catalog-first stays the default (structured artist/title/album, skip yt-dlp when it hits). If that batch comes back and the matcher rejects everything, we now fall through to ytsearch once. Empty catalog already did that inside search(). Remixes that only exist as videos should show up again without checking yt-dlp on every official-song search.

Cookie-strip: It was intentional for the premium qualities. Dropping cookies on the first retry made 774/141 disappear, so a probe that ranked opus 256 then fetched anonymous 160. I agree that this way it was going to cause issues. Changed so cookies stay for attempts 1 and 2 (web_music) but the last retry is muxed best with cookies dropped so expired/bot-flagged cookies don’t poison the walk. Import still probes the file on disk, so we don’t mistake a 160 file as premium.

PR #1169, opened 2026-08-21
fix/dev findings

nick2000713:
@Nezreka, was reviewing my library v2 branch against dev and found 16 things.. ten of those are the overhaul's own fault and stay in my branch. the other six are already sitting on dev and have nothing to do with library v2, so i pulled them out here.. this branch is straight off dev, doesn't touch a single lib2 file and depends on nothing else of mine.

all six have tests, full suite passes (14515).

## expired download cleaner can delete files people favourited (L2-001)

this is the one i'd take first.. it's supposed to spare anything someone favourited, rated or put in a playlist, but every db call it uses for that was fail open. write errors returned 0, read errors returned {}, and if one user failed halfway through a sweep it still stamped itself as a clean run..

all three look identical to the cleaner: nobody likes any of this.. and that's exactly when it starts deleting. so a db hiccup or a plex that's briefly down is enough to lose files.

storage calls raise now, the sweep records whether every expected server and user really got stored, and the cleaner won't delete without a complete fresh snapshot.. installs without a media server still work normally, that case gets recorded explicitly so it doesn't block forever.

## atomic album publish reports half albums as done (L2-002)

if file 3 of 10 fails, the first two stayed live and the batch was still marked Complete.. album half in the library, half in staging, and no retryable task left for the rest.

failed db repoints were only logged too, so the db kept pointing at a staging path the same publish deleted right after.. it's all or nothing now, any failure rolls everything back and the batch doesn't complete.

## a rate limit overwrites your manual matches (L2-005)

`honor_stored_match` returned False when the fetch failed.. and False means "no stored id, go search by name" to the worker. so one hiccup and the id you picked by hand gets replaced with whatever the fuzzy search came back with.

three states now, and an id only gets released by an explicit re match.. the failure goes in as `error` with a timestamp so backoff actually works.

## metasync `since` is compared as text (L2-010)

sqlite writes `updated_at` with a space, the api takes ISO with a T, and T sorts after space.. so "give me everything since midnight" returns nothing from today. with a cheerful 200 and an empty list, so it just looks like there were no changes.

normalised to UTC and compared through `julianday()` now.

## metasync incremental has holes (L2-011)

soul ids, canonical claims and album year updates all change the exported payload without touching `updated_at`.. and an album carries its artist's name while a track carries the album title, so renaming an artist rewrites payloads that then never get re exported.

those writers touch the timestamp now, and the filter looks at the parent rows too.

## soul_id_path never gets filled (L2-014)

additive column, so old rows are NULL and nothing ever fills them.. checked the history, the algorithm marker landed in march and the column in august, so basically every existing install has empty paths.

didn't want to regenerate the ids since peers already traded claims on them.. so there's a migration that recomputes each path to prove it, and writes `unknown` where it can't.

## unrelated

`pytest -n 8` is currently useless.. the xdist workers race in `video_database._ensure_columns` and you get hundreds of fake errors in suites that have nothing to do with any of this. serial runs are fine.

### DIFF
--- a/core/amazon_worker.py
+++ b/core/amazon_worker.py
@@ -8,7 +8,7 @@ from utils.logging_config import get_logger
 from database.music_database import MusicDatabase
 from core.amazon_client import AmazonClient
 from core.worker_utils import idle_backoff_seconds, interruptible_sleep, set_album_api_track_count
-from core.enrichment.manual_match_honoring import honor_stored_match
+from core.enrichment.manual_match_honoring import MATCHED, honor_stored_match
 from core.amazon_outage import is_source_outage, next_poll_delay_seconds
 
 logger = get_logger("amazon_worker")
@@ -390,14 +390,22 @@ class AmazonWorker:
         self._update_track(track_id, api_data, stored_id)
 
     def _process_album(self, album_id: int, album_name: str, artist_name: str, item: Dict[str, Any]):
-        if honor_stored_match(
+        _stored = honor_stored_match(
             db=self.db, entity_table='albums', entity_id=album_id,
             id_column='amazon_id',
             client_fetch_fn=lambda asin: self.client.get_album(asin, include_tracks=False),
             on_match_fn=self._refresh_album_via_stored_id,
+            mark_status_fn=self._mark_status,
+            status_column='amazon_match_status',
             log_prefix='Amazon',
-        ):
-            self.stats['matched'] += 1
+        )
+        if _stored:
+            # L2-005: a stored ID the source could not confirm right now is
+            # NOT released to a fuzzy name search below — a transient provider
+            # failure is not evidence that the ID is wrong, and searching
+            # overwrote deliberately chosen matches with whatever came back.
+            if _stored == MATCHED:
+                self.stats['matched'] += 1
             return
         # honor_stored_match also returns False when the stored id failed to
         # re-fetch (transient error / rate limit). Don't fall through to a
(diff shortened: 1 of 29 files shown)


nick2000713:
pushed one more thing.. this also fixes what #1103 reported, so that one can be closed in favour of this.

`_write_default_quality_profile_row()` wrote six of the fourteen columns on the `is_default` row, so `POST /api/quality-profile` took the other eight and dropped them without a word while `GET` hands back all fourteen. read, edit one field, post it back, and those eight could never change.

the fix in #1103 is to write all fourteen unconditionally, and that is where it goes wrong. the settings page posts only the six ladder fields and a preset posts five, so every key that is not in the body gets coerced to a default and written. i tried it against a real db: lossy copy, acoustid strictness, deep verify and replace lower quality all flip to off on an ordinary save, and the ladder editor autosaves, so reordering your formats is enough. silent too.. the pipeline only falls back to config when the key is absent, and after the wipe it is present and `False`.

those eight belong to the global config and reach the row through `sync_default_quality_profile_from_config()`. that is why they were left out. it just was not written down anywhere, which is why it read like an oversight.

so here the update is assembled from the ladder columns plus whichever bundle columns the caller actually named. presence decides, not truthiness, a field can go off as well as on, a six field body leaves the rest alone, and posting a full `GET` payload back is a no op. coercion goes through `_quality_profile_bundle_params` now, same helper create and update use, so the hand written copy is gone rather than extended.

while i was in there i also caught the ui half of the same area. editing one of those eight controls while previewing a non default profile does nothing and says nothing. the ladder controls already refuse and raise the editing banner, but the eight are plain config keys that go through the whole page save, and that one substitutes the real default's values back in for exactly those keys before sending. correct, and completely silent.. you toggle it, switch profiles, come back, and it reads Off again with nothing having told you why. the page autosave now raises the same banner when a profile captured control is edited against somebody else's profile. the save is untouched, the substitution already makes it a no op. the banner said "changes on the left", which is the ladder tiles, but six of the eight live in other sections, so it names the page now.

five backend tests and a set of ui ones. the backend five fail in both directions.. one against dev today for the original bug, three against the #1103 patch for the reset. the existing suite caught neither, `tests/quality` was 390 green with that patch applied.

full serial run here is 14576 passed. the two reds in `tests/test_mirrored_owned_counts.py` are not mine, they fail on plain dev at 8c2feebf0 the same way.

Nezreka:
Friggin beautiful. You are a rockstar and I can't tell you how much I appreciate you taking the time to make a side PR with these fixes you found. Everything seems great, a couple tiny things i'll resolve after I merge but you are killing it, man. thank you.

Issue #868, opened 2026-06-13
[BUG]  library view missing albums / different albums shown in Standard vs. Enhanced View


trackhacs:
i have the problem that the library view only shows subsets of the artist, compared to the available albums via jellyfin. there are also different albums shown depending if the library view set to "standard" or to "enhanced".  

deep sync with jellyfin did not repair this behavior. 

Example:

 

<img width="1236" height="841" alt="Image" src="https://github.com/user-attachments/assets/42b9cfa8-3511-410d-b391-b5e27578b156" />

####################################################################################
vs.
####################################################################################

<img width="1198" height="788" alt="Image" src="https://github.com/user-attachments/assets/4536851f-c6fc-40f5-a6e1-2594d01287f5" />

Nezreka:
Nice catch. I have the fix ready for this on dev and will push main soon. What you can do now though is rematch this library artist to the correct artist. In the photo you sent there are green buttons that say like: 'Musicbrainz: matched' and 'Deezer: matched'. Click on each of those and manually match them to the correct Rone artist.

Issue #586, opened 2026-05-14
Aliases of artists still not being matched


nextfinish:
Hello, following up on issue #442, I am still having issues with aliases of artists.

Example of a failed download:
AcoustID verification failed: Audio mismatch: file identified as 'Ballet Suite no. 2: III. Polka' by 'Русская филармония, Дмитрий Яблонский', expected 'Ballet Suite No. 2: III. Polka (Suite for Jazz Orchestra No. 1)' by 'Dmitry Yablonsky' (title=100%, artist=4%) — expected artist not found in any AcoustID recording

For context, Дмитрий Яблонский is the russian spelling of Dmitry Yablonsky, so it should be matching.

Moreover, it would be nice to have a Tool to merge aliases (or even simply different capitalizations of artist names).

Nezreka:
Okay this should be much better now on dev branch. I like your idea of a tool though, will look into that.

PR #1185, opened 2026-08-25
Cross-script identity: stop quarantining correct downloads, and make the scan agree with the download

nick2000713:
# AcoustID: correct downloads were being reported as wrong downloads

I found this while working on Library V2, not while looking for AcoustID bugs.

What kept happening: a file passed verification at download time, and later the
library scan called the same file a wrong download. Nothing changed in between.
Scan again and it said something else, first "Mismatch", then "unverified".

I chased this for a long time and every explanation I had was wrong. It looked
like a scanner bug, then a MusicBrainz bug, then a caching problem, then a
regression from my own earlier fix. It was all of them stacked, which is why
fixing any single one changed nothing I could see.

The trigger in my library, both files perfectly correct:

```
Wrong download: "Apetitan" is actually "APETITAN"
Expected "Apetitan" by Sawano Hiroyuki, but the audio fingerprint matches
"APETITAN" by 澤野弘之 (fingerprint 100%, title 100%, artist 0%)
```

8 commits, each a separate defect with its own tests.

### 1. A different writing system is not evidence

Romaji against kanji scores 0.00 whether or not it is the same person.
`evaluate` compared that to `CLEAR_MISMATCH_THRESHOLD` and called it a clear
wrong song. Now a dimension written in two different scripts counts as unknown,
never as failed, and can never be the reason for a FAIL. Same-script
disagreement is still evidence, so a wrong artist or song still fails (four
tests, including two kanji artists against each other).

This also drops the `fingerprint >= 0.95` floor for cross-script titles. Below
it, an ordinary 0.90 match on a Japanese title was getting quarantined.

Known gap, written down in a comment and a test: if title *and* artist are both
unreadable, nothing agrees and nothing disagrees, so a real wrong download with
a non-Latin recording cannot be reported here. I built a fingerprint bar for
that and reverted it. The score says nothing about the names, so all it does is
quarantine the correct files again.

### 2. A timeout is not an answer

This is what made it intermittent, and it took me longest to find.

`MusicBrainzClient.search_artist` returns `[]` for a timeout, a 429 and a 503.
It returns the same `[]` for "MusicBrainz knows nobody by that name". So the
alias lookup cached an outage as "this artist has no aliases" and blocked every
retry for the cache TTL. A bulk scan is exactly what trips the rate limit, so
one sweep could kill the romaji/kanji bridge for a month. My database had such
a row.

`search_artist` and `get_artist` now take `raise_on_error`, off by default so
other callers keep their fail-soft list. Only a real answer gets written down,
and rows from the old code get one retry, which heals the poisoned ones already
out there.

### 3. Use the MBID instead of guessing from the name

`lookup_artist_aliases` searched by name and pulled aliases from whatever
entity passed a trust gate. The gate's own comment names the artist it loses
on: for "Sawano Hiroyuki" a decoy leads at 0.82 while the real 澤野弘之 sorts
last. There is nothing to guess once `artists.musicbrainz_id` is set, so a new
tier in front of the name search uses it.

What a lookup learns now also stays on the artist row instead of only in a
cache keyed by whatever spelling the caller happened to pass, guarded by
`source_id_conflict` like the enrichment workers.

### 4. Cross-script artists could never be matched automatically

`match_artist` scores 60% name similarity plus 40% MB relevance, so a name in
another script caps at 40 against a gate of 70. Unmatchable by design, which is
why the MBID point 3 needs was never there.

When the name path finds nobody, candidates are now checked against the albums
the user owns (`catalog_overlap_score`, same signal as #868). Strict: MB score
>= 90 and at least one owned album. No owned albums, no match.

It runs last, and that is the safety argument. An earlier version ran it before
the confidence gate and beat a 95-scoring same-script candidate with an
80-scoring one, on a shared album title as common as "Home".

Also #586 for this function: it asks non-strict when strict returns nothing.
That was fixed for `lookup_artist_aliases` and never came back for this one.

### 5. MusicBrainz search needed a worker it does not need

`_search_service` raised "MusicBrainz worker not initialized" whenever
`mb_worker` was None, which it is if the worker's init failed. MusicBrainz needs
no credentials and there is already a shared, rate-limited client. The worker's
service is still preferred when it exists so both share one limiter.

### 6. A scan could take a "verified" away

The scan reads the standing from the file tag and writes it to
`tracks.verification_status`. With the tag missing (import write did not stick,
file copied, later retag) it saw an untagged file, applied the rule for one, and
downgraded a row the catalogue said was verified. Then it stamped that answer
onto the tag.

Before point 1 this was unreachable, because the decision was FAIL and FAIL
leaves the status alone. That is the "Mismatch" step. Turning that FAIL into an
honest SKIP walked the file into the downgrade, which is why the order I saw the
symptoms in reads as if my own fix caused it.

Either record saying "verified" now means verified, and a missing tag gets
healed instead of overwritten.

### 7. Scan and download are one path now

The final decision was already shared. Everything before it was not: the scan
had its own availability check, lookup, confidence floor and alias wiring, and
never enriched title-less recordings from MusicBrainz the way the download does.
So a fingerprint hit whose title AcoustID does not carry got dropped by one path
and judged by the other.

`_scan_file` calls `AcoustIDVerification.verify_audio_file` now. What is left in
the scanner is what only the scan can answer: which files, which artist value to
compare against, and what to do with a verdict.

Two behaviour changes fall out, both on purpose:

* An unexpected exception is `ERROR`, not `SKIP`. It was never a statement about
  the file, but read as one it became a verdict: the scan turns SKIP on an
  untagged file into "unverified", the download turns it into a rejection under
  `require_verified`.
* `scan()` refuses to start if the client is not usable. Per file that answers
  SKIP, so a missing API key would walk the whole library into the review queue
  and report a clean run.

Covered by a test that runs eight cases through both entry points and asserts
the same verdict, plus a structural guard so the scanner cannot grow its own
copy of the decision logic again.

### The last commit

I had the branch reviewed and it found five more things, all real. One is
serious: the `[]`-on-timeout in point 2 lives in the client, so the
service-level fix could not fire at all. The rest are a failed fallback search
still reaching the negative cache, MusicBrainz result order used as a tie-break
in point 4, a display name treated as an identity (`LIMIT 1` over rows sharing a
name), and `int()` on an artist id that is TEXT after the `artists_new`
migration, which raises on Jellyfin GUIDs.

One thing I fixed on the side: four tests in `test_acoustid_scanner.py` were
calling MusicBrainz over HTTP for real, 35 seconds each and a different answer
when rate-limited. An autouse fixture stubs it, 44s down to 0.3s.

Happy to split this into smaller PRs if you would rather take it in pieces.

### DIFF
--- a/core/acoustid_verification.py
+++ b/core/acoustid_verification.py
@@ -241,7 +241,8 @@ class AcoustIDVerification:
         audio_file_path: str,
         expected_track_name: str,
         expected_artist_name: str,
-        context: Optional[Dict[str, Any]] = None
+        context: Optional[Dict[str, Any]] = None,
+        min_score: Optional[float] = None,
     ) -> Tuple[VerificationResult, str]:
         """
         Verify that an audio file matches expected track metadata.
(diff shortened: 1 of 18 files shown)


Nezreka:
this is a seriously good pr. the five-defects-masking-each-other story checked out top to bottom - i ran your suites plus the full blast radius on a merge against current dev (2300+ tests green), and reverting your fixes fails your tests like they should. the known-gap comments in evaluate are exactly how i want tradeoffs written down.

one thing i found: the raise_on_error plumbing on the real client isn't covered - the resilience tests stub a client that already honors the flag, so deleting the raise stays green. not blocking, i'll add the small client-seam test on dev after merge.

merging. thanks for the scanner test stub too, that kills a real ci flake.

PR #1188, opened 2026-08-26
fix(acoustid): duration-aware candidate selection and cluster disambiguation

shkarlsson:
### Summary of Changes

Fixes an issue where AcoustID fingerprint lookups for tracks with multiple candidate mixes/versions in a single AcoustID cluster (e.g. *Radio Version [216s]* vs *3 AM Mix [416s]*) falsely failed verification as 'Wrong download'.

### Key Changes
1. **Preserve Recording Durations in AcoustID Client (`core/acoustid_client.py`)**:
   - Query AcoustID with `parse=False` so that candidate recording objects retain their `duration` metadata.
2. **Duration- & Subtitle-Aware Candidate Selection (`core/matching/audio_verification.py`)**:
   - Scored candidates in `_find_best_title_artist_match()` using audio duration matching ($\pm 60 or $\pm 35\%$ tolerance) and verbatim subtitle matching before parenthetical stripping.
   - Wired `expected_duration_s` through `evaluate()`.
3. **Scanner & Verifier Wiring (`core/repair_jobs/acoustid_scanner.py` & `core/acoustid_verification.py`)**:
   - Passed `file_duration_s` from disk/context into `evaluate()` so duration filtering actively runs during library scans and import verification.
4. **Unit Tests (`tests/matching/test_audio_verification_core.py`)**:
   - Added unit tests for duration-based candidate disambiguation and tie-breaking on subtitle variations.

### Verification
- Full test suite passing (255+ tests).
- Verified against real test corpus:
  - Instrumental vs. Vocal mismatches $\rightarrow$ strictly rejected (`FAIL`).
  - True wrong downloads (P2P renames / wrong artist) $\rightarrow$ strictly rejected (`FAIL`).
  - Extended/Mix cluster candidates $\rightarrow$ correctly disambiguated to matching length (`PASS`).

### DIFF
--- a/core/acoustid_client.py
+++ b/core/acoustid_client.py
@@ -386,26 +386,59 @@ class AcoustIDClient:
             seen_mbids = set()
             best_score = 0.0
 
-            for result in acoustid.match(self.api_key, audio_file, parse=True):
-                # match() with parse=True returns (score, recording_id, title, artist)
-                if not isinstance(result, tuple) or len(result) < 2:
-                    logger.warning(f"Unexpected result format: {result}")
-                    continue
+            # Query AcoustID with parse=False when possible to retain full recording metadata (including duration)
+            try:
+                lookup_res = acoustid.match(self.api_key, audio_file, parse=False)
+            except TypeError:
+                lookup_res = acoustid.match(self.api_key, audio_file)
 
-                score = result[0]
-                recording_id = result[1]
-                title = result[2] if len(result) > 2 else None
-                artist = result[3] if len(result) > 3 else None
+            if isinstance(lookup_res, dict) and lookup_res.get('status') == 'ok':
+                for result in lookup_res.get('results', []):
+                    score = float(result.get('score', 0.0))
+                    if score > best_score:
+                        best_score = score
+                    for recording in result.get('recordings', []):
+                        recording_id = recording.get('id')
+                        title = recording.get('title')
+                        duration = recording.get('duration')
+                        artists = recording.get('artists')
+                        if artists:
+                            artist = "".join([a.get('name', '') + a.get('joinphrase', '') for a in artists])
+                        else:
+                            artist = None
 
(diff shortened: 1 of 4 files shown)


Nezreka:
Nice! I'll be able to review this soon, going on a little vacation and be back Monday. Thanks for your help!

Nezreka:
nice catch, and the fix is the right shape. duration comes in as a tie-breaker rather than a new threshold, so title and artist similarity still decide and the pass/fail gates don't move. reusing duration_mismatches_strongly instead of inventing another tolerance was the right call too. 1115 tests pass across acoustid and matching.

two small things before i merge.

ruff fails on this. clean on dev, one error here:

    S110 try-except-pass detected, consider logging the exception
    core/acoustid_verification.py:386

a logger.debug in the handler does it. worth fixing properly rather than noqa, silent excepts around file probing are how the docker path bugs stayed invisible for months.

the other one is subtler and i don't think you meant it. parse=False skips the status check.

acoustid answers HTTP 200 with {"status": "error"} for a bad api key or a quota problem. _api_request hands that json straight back, and it's parse_lookup_result that raises WebServiceError on a non-ok status. with parse=True that lands in your outer except and the client returns status error with a log line. with parse=False nothing checks it, so the error dict misses the status == 'ok' branch, falls into the iterable branch, iterates the dict's keys, finds no tuples and reports "No AcoustID matches found".

so a wrong api key now reads as every file having no fingerprint match, with nothing in the log. something like:

    if isinstance(lookup_res, dict) and lookup_res.get('status') != 'ok':
        msg = (lookup_res.get('error') or {}).get('message') or 'unknown error'
        logger.error(f"AcoustID lookup failed: {msg}")
        return {'status': 'error', 'recordings': [], 'error': msg}

before the iterable fallback.

neither touches your actual logic. sort those two and i'll merge it.

PR #1190, opened 2026-08-26
fix(duplicates): preserve intentional lossy companions

nick2000713:
## What was the problem?

When lossy copy is enabled, SoulSync intentionally creates an MP3, Opus or AAC/M4A copy next to the original lossless file. The Duplicate Detector already knew that this is an intentional pair and did not report it as a duplicate.

The filesystem Duplicate Cleaner did not use the same exception. Because both files have the same folder and filename stem, it could detect the generated lossy copy as a normal cross-format duplicate and move it into the `.deleted` quarantine folder.

So in practice the lossy copy feature could create the file correctly, but a later duplicate-cleaner run could remove it again.

## What does this PR change?

- Adds one shared rule for detecting intentional lossless + lossy companion pairs.
- Uses this rule in both the Duplicate Detector and the filesystem Duplicate Cleaner, so the two systems no longer behave differently.
- Reads the enabled codec from the global lossy-copy setting and also from enabled quality profiles.
- Supports the currently available MP3, Opus and AAC/M4A companion formats.
- Keeps the intentional lossy copy next to its lossless source instead of moving it to `.deleted`.
- Still detects and quarantines real same-stem cross-format duplicates. For example, a FLAC + OGG pair is still cleaned when only MP3 is configured as the lossy-copy format.
- Keeps the old behaviour when lossy copy is disabled, so normal duplicate detection is not weakened globally.

## Tests

I added tests for the important cases:

- Global MP3 lossy copy enabled: FLAC and MP3 are both kept.
- Lossy copy enabled through a quality profile: FLAC and Opus are both kept.
- Real cross-format duplicate: FLAC is kept and OGG is still quarantined.
- Lossy copy disabled: the previous duplicate-cleaner behaviour stays unchanged.

In total, 83 focused duplicate, quarantine and repair tests pass. Python compilation and `git diff --check` also pass.

I also started the complete Python test suite. The failures seen there come from unrelated network/worker tests trying to access services such as MusicBrainz, Last.fm and Genius in the isolated test environment.

### DIFF
--- a/core/library/duplicate_cleaner.py
+++ b/core/library/duplicate_cleaner.py
@@ -9,7 +9,9 @@ import logging
 
 from core.settings import config_manager
 from core.runtime_state import add_activity_item
+from core.library.duplicate_rules import is_lossy_companion_pair, lossy_companion_exts
 from core.repair_jobs.base import skip_deleted_quarantine
+from database.music_database import get_database
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


Nezreka:
I'll check this out soon! Heading out of town for a few days

PR #1206, opened 2026-08-28
Fix Discogs discography classifier mislabeling compilations and videos as "Album"

### LINKED ISSUE #1205: Discogs discography classifier mislabels compilations and concert-video releases as "Album"
## Summary

When browsing an artist's discography via Discogs, best-of compilations and concert Blu-rays/DVDs show up mixed in with genuine studio albums, all labeled as plain "Album."

## Root cause

`Album.from_discogs_release()` / `get_artist_albums()` in `core/discogs_client.py` classify anything they can't confidently match to a keyword as generic `album`. Two separate bugs make that happen far more than it should:

1. **Compilations never match.** The classifier checks for the substring `'compilation'`, but the `/artists/{id}/releases` endpoint `get_artist_albums()` actually calls returns Discogs' **abbreviated** format code `"Comp"` — never the full word. Every compilation silently falls through to the `album` default.
2. **No filtering by medium at all.** There's no handling for non-audio formats (Blu-ray, DVD-V, Laserdisc, CDV, professional broadcast-tape formats like Umatic/Betacam) — these land in the generic `album` bucket alongside real studio albums, since neither the keyword checks nor the track-count fallback catch them.

## Steps to reproduce

1. Use Discogs as (or one of) the metadata source(s).
2. Open an artist with compilations and/or concert video releases in their Discogs catalog (e.g. Naoya Matsuoka, Discogs artist ID `1158913` — `Drive Best!` is tagged `CD, Comp, RM`; `Live At Montreux Festival 1980` is tagged `Blu-ray`).
3. View the artist's discography.

**Expected:** compilations are recognizable as such (or at least filterable via the existing Compilations toggle); concert videos don't appear as downloadable albums at all.
**Actual:** both show up as plain "Album," indistinguishable from a real studio album.

## Impact

Verified against real, live Discogs data across four artists of very different shape and scale (not just the reproduction case): a 205-release catalog, a 856-release catalog (150-item production window), and two full-catalog stress tests at 3,825 and 10,567 raw releases. On the two full-catalog runs, 18 and 29 releases respectively were confirmed (by hand) as genuinely non-audio and currently mislabeled as `album`.

I have a fix (three real design iterations, each forced by a concrete release that broke a simpler version — details in the PR) and will open a PR against `dev` referencing this issue. Thanks for SoulSync!

RiceTeaPrince:
Fixes #1205

## Problem

`Album.from_discogs_release()` / `get_artist_albums()`
(`core/discogs_client.py`) classify every Discogs release they can't
confidently match to a keyword as generic `"album"`. Two distinct bugs make
this happen far more than it should, so an artist's discography view mixes
genuine studio albums with best-of compilations and concert Blu-rays/DVDs,
shown as if they were ordinary albums:

- **Compilations never match.** The classifier checks for the substring
  `'compilation'`, but the `/artists/{id}/releases` endpoint
  `get_artist_albums()` actually calls returns Discogs' **abbreviated**
  format code `"Comp"` — never the full word. Every compilation silently
  falls through to the `else: album_type = 'album'` default.
- **No filtering by medium at all.** There's no handling for non-audio
  formats (Blu-ray, DVD-V, Laserdisc, CDV, professional broadcast tape
  formats like Umatic/Betacam) — these land in the generic `album` bucket
  alongside real studio albums, since neither the keyword checks nor the
  track-count fallback catch them.

## Fix

Two independent, additive changes to `core/discogs_client.py`:

1. **Match `Comp` alongside `compilation`**, and fold `ep` into `single`
   (Discogs' own artist pages have no separate EP facet — "Singles & EPs"
   is one bucket there; `Album.album_type` keeps its existing four-value
   vocabulary everywhere else, this merge is scoped to the Discogs
   converter only).
2. **Exclude non-audio-format releases from `get_artist_albums()`'s
   results**, before classification, via a new `_is_non_audio_discogs_release()`
   helper — rather than inventing a new `album_type` value the webui's
   discography views (Albums/EPs/Singles buckets, no Video bucket) have
   nowhere to surface.

Change 2 went through three real design iterations, each forced by a
concrete release that broke the simpler version, not by caution alone:

- A blacklist of known video formats missed a Michael Jackson promo on
  `Umatic` (professional broadcast tape) — the video/broadcast-format tail
  is long and unenumerable, unlike the small, stable set of audio carrier
  formats.
- A naive whitelist (flip it: exclude unless a known-audio token is found
  anywhere) fixed that but wrongly un-excluded a genuine Laserdisc release
  whose own disc size (`12"`) got misread as "a vinyl record is also here."
- The final version checks each `+`-separated bundled item
  (Discogs joins e.g. `Blu-ray + 2xCD`) by its own leading format name
  rather than pooling every token across the whole string — this also
  fixed an unrelated bug where the blacklist version dropped real mixed
  CD+DVD releases (audio CD with a bonus DVD) wholesale.

`Promo`/`Unofficial`/`Ltd`/etc. are deliberately **not** treated as
exclusion signals anywhere in this fix — they're Discogs *modifiers* on a
release that's still legitimately album/compilation/single, not a *type*,
and excluding on them would override Discogs' own type judgment instead of
trusting it (concrete before/after examples in the linked report below,
including three `Comp`+`Promo` releases the modifier-based approach would
have dropped from the discography entirely).

**Deliberately out of scope, documented as a follow-up**: Discogs `master`
entries carry no `format` field at all on the endpoint this code calls, so
a compilation/video release that's dedup-preferred as a master still lands
on `album`, unaffected by either change. A fix sketch (fetch
`/masters/{id}` → `main_release` → `/releases/{id}`, classify that instead)
is tested and works, but costs 2 extra rate-limited API calls per master —
not viable inline on a page load, would need to be a background enrichment
step. Not included here.

Full root-cause detail, every design iteration, and the complete real-data
verification (four artists, tables of before/after per release) are written
up in #1205.

## Tests

`tests/test_discogs_discography_classifier.py` (new file, 23 tests):

- Change 1 — `Comp` abbreviation recognized, full-word `compilation` still
  works, `ep` folds into `single`, explicit `single`/`album` format strings
  unaffected, the collapsed `<=6`-track fallback, and `Promo` confirmed as
  a non-exclusion signal (a `Comp, Promo` release still classifies as
  compilation).
- Change 2 — every release type from the real verification data:
  pure-video (Blu-ray, DVD-V, CDV) excluded; the Laserdisc-with-its-own-12"
  collision that broke the naive whitelist still correctly excluded; the
  exact Umatic-tagged release that broke the blacklist (carrying a
  `Single` type marker alongside the video format) still correctly
  excluded; a `Blu-ray + 2xCD` bundle and a `CD, Album + DVD` bundle both
  correctly kept (real audio content present); a `DVD-D` data disc of MP3s
  resolving to audio; plain vinyl and vinyl-in-a-box kept; a master with no
  format data left alone (not falsely excluded); and a fully unrecognized
  format string left unguessed.

## Verification

Checked against real, live Discogs API data across four artists of very
different shape and scale — not just the originally reported catalog:

| Artist | Raw releases | Candidates after existing role/dedup filter | Result under this fix |
|---|---|---|---|
| Naoya Matsuoka (id `1158913`) | 205 | 53 | 46 albums / 3 singles / 4 excluded (non-audio) |
| Tatsuro Yamashita (id `119485`), production-realistic 150-item window | 150 | 71 | 40 albums / 4 singles / 0 excluded in this slice |
| Metallica (id `18839`), full catalog | 3,825 | 461 | 416 albums / 18 singles / 18 excluded, 0 false positives (checked by hand) |
| Michael Jackson (id `15885`), full catalog — the artist whose `Umatic` release drove the whitelist redesign | 10,567 | 446 | 354 albums / 31 singles / 29 excluded, 0 false positives (checked by hand) |

Every exclusion across the two full-catalog runs (47 releases) was
individually reviewed, not just aggregate counts. Known, deliberately
undocumented-as-fixed gaps found along the way (each confirmed against real
data, not theoretical): a release identifiable as non-album only via
free-text `notes` the list endpoint doesn't return; a Discogs list-endpoint
response with a truncated, dangling `format` string on one release out of
10,567 (a Discogs data-completeness issue, not something parseable-around
on this end); and Metallica's active-tour discography making
`get_artist_albums()`'s 150-most-recent-releases fetch window return
zero classic studio albums for a sufficiently prolific/active artist — a
separate, unrelated bug flagged for its own issue, not fixed here.

Ran this repo's actual CI gates locally (Python 3.11.15 via `uv`,
`requirements-dev.txt` pinned versions, `webui/` deps via `npm ci`):

- `ruff check .` — clean
- `compileall api core database services scripts web_server.py wsgi.py beatport_unified_scraper.py` — clean
- `pytest` — 15170 passed, 2 skipped, 7 deselected, 0 failed (15147 baseline + this PR's 23 new tests)

### DIFF
--- a/core/discogs_client.py
+++ b/core/discogs_client.py
@@ -323,18 +323,25 @@ class Album:
         else:
             format_str = str(raw_format).lower()
 
-        if 'single' in descriptions or 'single' in format_name or 'single' in format_str:
+        if ('single' in descriptions or 'single' in format_name or 'single' in format_str
+                # Discogs has no separate EP facet on its own artist pages
+                # ("Singles & EPs" is one bucket there) -- treat it as
+                # 'single' here too rather than inventing a split the
+                # source doesn't make.
+                or 'ep' in descriptions or ', ep' in format_str or format_str.endswith('ep')):
             album_type = 'single'
-        elif 'ep' in descriptions or ', ep' in format_str or format_str.endswith('ep'):
-            album_type = 'ep'
-        elif 'compilation' in descriptions or 'compilation' in format_str or 'compilation' in (release_data.get('type', '') or '').lower():
+        elif ('compilation' in descriptions or 'compilation' in format_str
+                # Discogs' artist/search-release endpoints use the
+                # abbreviated code "Comp", not the full word -- this was
+                # previously never matched, so every compilation fell
+                # through to the generic 'album' branch below.
+                or 'comp' in descriptions or 'comp' in format_str
+                or 'compilation' in (release_data.get('type', '') or '').lower()):
             album_type = 'compilation'
         elif 'lp' in descriptions or 'lp' in format_str or 'album' in descriptions or 'album' in format_str:
             album_type = 'album'
-        elif total_tracks <= 3 and total_tracks > 0:
-            album_type = 'single'
         elif total_tracks <= 6 and total_tracks > 0:
-            album_type = 'ep'
+            album_type = 'single'
         else:
             album_type = 'album'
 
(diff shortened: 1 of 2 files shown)


Nezreka:
good work, both bugs are real and you found the actual cause of each.

the Comp one is on me. that endpoint has been sending "Comp" the whole time and the check wanted the full word, so every compilation became an album.

the non-audio helper is the good part. splitting on "+" so "Blu-ray + 2xCD" keeps its cd, not dropping an album with a bonus dvd, leaving masters alone. the laserdisc whose own 12" size reads as vinyl is nasty, i wouldn't have caught it.

checked it properly. your 23 pass, 67 existing discogs pass, 299 across discography, ruff clean. reverted each fix on its own to make sure your tests catch them, they do.

one thing before i merge. folding ep into single isn't part of #1205 and it breaks the EPs tab on artist pages:

    core/metadata/discography.py:544
    if album_type == 'ep':

discogs can't produce 'ep' anymore so that tab goes empty and the eps land under singles. deezer and itunes still make eps, so discogs would be the only source disagreeing on the same release.

keep the Comp match and the exclusion as they are, put the ep branch back, restore the <=3 single / <=6 ep fallback. small edit.

none of my tests caught it, so that's a gap on my side.

fix that and i'll merge.

PR #1220, opened 2026-09-02
feat(settings): move YouTube auth into Connections

ht5161143:
## Summary

YouTube Music polling and yt-dlp downloads share the same cookie auth, but the cookie fields only rendered when YouTube was the active download source. 

That basically means the user has to change it's download source to "YouTube only", paste the cookies, and then revert to the previous download source. 

This PR changes this by moving the youtube auth part in the "Connection" tab of the settings

Before 👇 

Screenshot 2026-09-02 at 15 50 40

After 👇 

Screenshot 2026-09-02 at 17 23 45

Screenshot 2026-09-02 at 16 53 44

### DIFF
--- a/api/source_playlists.py
+++ b/api/source_playlists.py
@@ -4657,7 +4657,7 @@ def get_ytmusic_playlists():
     from core.ytmusic_library import fetch_library_playlists, fetch_liked_music_row, library_playlists_to_rows
     auth = _ytmusic_auth_headers()
     if not auth:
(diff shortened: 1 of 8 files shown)


ht5161143:
Do you think I should show the cookies.txt fields in both views as it is done for Deezer? 

Screenshot 2026-09-02 at 18 05 53

Screenshot 2026-09-02 at 18 06 03

PR #1262, opened 2026-09-17
fix(soulseek): improve download selection, retries, and album matching

mandos21:
## Summary

This PR is a focused set of Soulseek downloading fixes aimed at making automatic downloads choose stronger candidates, attempt to find faster candidates safely, and avoid incorrectly treating a nearby file as the requested track.

In local/manual testing, the updated selection behavior has shown anecdotal evidence of materially better performance: previously troublesome tracks have often found and completed from stronger peers much more quickly. This is deliberately described as anecdotal rather than a benchmark; Soulseek peer availability and transfer performance are inherently variable.

The changes retain conservative correctness gates and always preserve a last-resort path: automatic retries do not turn a slow but viable download into a missing track.

## What changes for downloads

- Search collection now merges evolving slskd response snapshots by peer and file instead of dropping late files or stopping after 30 responses. It keeps polling an active search through a quiet period, while still exiting promptly once slskd reports a terminal state.
- Soulseek candidates are ranked within small confidence bands using observed peer throughput, free slots, queue depth, active occupancy, advertised upload speed, and quality. This keeps correctness/confidence primary while allowing similarly credible fast peers to win over correct but slow options.
- Candidate order interleaves peers within a band so one uploader with many hits does not monopolize the retry walk.
- slskd's current `hasFreeUploadSlot` response field is respected, with compatibility for the older numeric field.
- Self-titled album searches avoid redundant `Artist Artist` queries and structured path handling recognizes common `Artist/(Year) Album/01 Track` layouts.

## Observed-speed fallback

- Adds an opt-in Soulseek setting, **Retry slow Soulseek downloads**, with a default minimum observed speed of **250 KB/s**. It is disabled by default.
- Measures actual byte progress over a rolling 30-second window and requires a second sustained low-speed period before switching, so transient dips and oscillating transfers are not abandoned.
- Retries cached, validated alternatives before issuing fresh searches. The previously accepted slow source is retained as the final fallback if all alternatives are exhausted.
- A replacement is only started after slskd cancellation is confirmed terminally. If cancellation is rejected or cannot be confirmed, SoulSync restores and exempts the original transfer rather than creating competing downloads.
- Album bundles use equivalent aggregate-throughput logic and only change folders when another validated folder can fulfill the remaining tracks. Completed tracks are retained across the switch.

## Album and track matching

- Adds target-aware filename/path interpretation for Soulseek files rather than trusting a generic filename parser. It recognizes leading and embedded track numbers, disc folders, artist/album prefixes, technical tags, and supported preferred versions.
- Album-folder selection now uses one-to-one expected-track coverage, preventing a single ambiguous result from satisfying multiple requested tracks and rejecting partial folders before bundle dispatch.
- Album-title similarity can discount distinct artist/year metadata outside the album span, improving structured-folder matching without rewriting or making unsafe assumptions about music titles.
- Eponymous releases, year-prefixed folders, compilation artist routing, and unknown compilation track positions are handled explicitly.
- Source reuse and private bundle staging now use the same identity gate, preventing a sibling track from being reused simply because its folder metadata scores well.

## Wishlist and import correctness

- Prevents false wishlist completion from fuzzy transfer-folder matches. A transfer-file completion must match the exact destination path recorded by the import pipeline; it no longer reconstructs a hard-coded `NN - Title.flac` filename.
- This works with custom import templates and unknown compilation track numbers, while refusing to mark a task complete when no recorded destination can establish identity.
- Compilation paths use each track's artist for `$artist` while retaining the shared album artist for `$albumartist`.

## Internal cleanup

- Consolidates the 500 KB/s observed-peer ranking threshold under a named constant, separate from the user-configurable retry floor.
- Removes duplicate album coverage calculations and dead timeout logging.
- Adds focused regression tests for custom import paths, unknown-number transfer matches, cancellation rejection/confirmation, active-search quiet periods, fractional track numbers, album switching, and slot parsing.

## Validation

- `ruff check --no-cache` on the changed Soulseek/download/import paths and tests
- `git diff --check`
- `124 passed` focused post-processing, identity, observed-speed, and download-master tests after merging latest `dev`
- `9 passed` selected album-bundle/search regression tests after merging latest `dev`

The broader album-pinning suite contains a pre-existing long-running test in this local environment, so it was not represented as a full-suite green result.


Nezreka:
Thanks for this, the observed speed idea is a good one and the tracker itself reads clean. A few things need fixing before I can merge though.

1. Search terminal state never fires. slskd reports search state as a flags string, "Completed, TimedOut" or "Completed, ResponseLimitReached". I pulled every search off my live slskd and not one of them is plain "Completed", so `_search_is_terminal` always returns False. Since the state is non empty the quiet period fallback is skipped too, and with the old 30 response early exit gone every search now runs the full timeout plus buffer (75s default). The payload has an `isComplete` bool, use that, or substring match like the transfer checks already do. Please add a test with "Completed, TimedOut".

2. The identity gate in `_match_filename_candidates` rejects files that are correct. I replayed `match_track` over every completed soulseek download in my own history (1675 rows, real filenames) and it rejects 232 of them, about 14%. Things like `13 - Eminem feat. Dina Rae - Superman.flac`, `03-fountains_of_wayne-stacys_mom.flac`, `0113 - Katy Perry - Peacock.flac`, `Artist_Album_01_Title.flac`, and any spotify title with a " - Remastered 2011" suffix. "7 rings" fails against `7 rings.flac` because the leading number regex eats the 7. This can't fail closed on a layout it doesn't recognize. Either make it advisory (reject only when the title clearly matches a different track on the album) or fail open when the engine score is already above threshold. Same gate is copied into source reuse and staging so those need the same change.

3. Bundle speed check has no state gate. The monitor path requires InProgress but the bundle path counts bytes as soon as slskd lists the transfer, so a folder sitting at "Queued, Remotely" for 60s is declared crawling, cancelled, and the peer gets stored at 0 bps for an hour. That's inside the 180s stall grace the bundle already has. Only accumulate bytes from InProgress transfers and skip observe_peer when nothing moved.

4. The new `artist_score < 0.65` hard reject breaks compilations and artist-less folders. Before it was weighted. With 100% track coverage, `Music\A Night at the Opera (1975) [FLAC]` scores 0, and so does every compilation layout I tried including `Music\Various Artists\Now 50`, because `_extract_album_artist` picks the most common per file artist (Kylie Minogue) instead of Various Artists. Put it back to weighted, or skip the gate for compilations and when coverage is 1.0.

5. Four existing tests fail on the branch: three in `tests/test_soulseek_album_fallback.py` (the stub doesn't take the new `deadline` kwarg) and `tests/downloads/test_post_processing_pool.py::test_search_work_still_goes_to_the_search_pool` (it counts 2 worker submits, now there are 3). All 12 pass on dev.

Happy to take another look once those are in.

mandos21:
Thanks for the detailed pass and apologies for the mistakes— all five are addressed and pushed, plus a few other things I found while reviewing again.

1. Search terminal state — _search_is_terminal splits on , and honors isComplete. Parametrized test covers "Completed, TimedOut", "Completed, ResponseLimitReached", and isComplete: true with a non-terminal state string. Also now consults state on zero-response searches, which were still burning the full timeout+buffer.
2. Identity gate — I ended up taking both of your options, one per path: per-track, the engine's 0.58 pre-filter means "fail open above threshold" is the same as never rejecting, so match_track now grades and never rejects; for album coverage, where there's no engine score, one-to-one assignment with ranked edges is the "only reject when it clearly matches a different track" form. Two things I did that you didn't ask for and should look at: (1) the track-prefix regex moved out of matching_engine into core/text/track_prefix.py so identity and base_title_of share it — 3-line change in the engine, additive, two internal callers; (2) the recovery step in _match_filename_candidates re-admits sub-threshold rows on exact identity + artist-in-path — it's the one place this PR accepts something the old code wouldn't, so it's the one I'd most like your eyes on.
3. Bundle speed gate — only InProgress bytes count, observe_peer skipped when nothing moved. Also found the inverse bug: _should_retry_task popped the tracker before the completion branch ran, so healthy per-track peers were never recorded and the ranking tier was unreachable. Fixed + regression test.
4. artist_score hard reject — replaced with coverage + album-title gates; artist is scored per path segment. Music/A Night at the Opera (1975) [FLAC] → 0.84, Music/Various Artists/Now 50 → 0.89, both above _ALBUM_PREFLIGHT_MIN_SCORE. The parent-directory-as-artist fallback is gone.
5. Tests — 3268 pass across tests/downloads, tests/imports, tests/wishlist, tests/matching, tests/test_soulseek*.

Replay: I ran match_track over 1,737 completed Soulseek downloads from my own history (618 peer paths, 1,119 bundle-staged). Accuracy went from 88.7% to 97.2% exact, with no regressions. The 2.8% left are now left to the confidence check.

Other changes since your review (all in the PR description): pending transfers are only cancelled when the folder was abandoned for crawling (an unresolved result is a resolver problem another peer would repeat); a folder whose transfers all fail is now replaced from the next ranked alternative; the bundle switch flow is split into three helpers and _poll_album_bundle_downloads always returns its detail dict; one peer_availability_key for per-track and preflight.

Also, if you'd prefer, I can split this into several smaller PRs.

PR #1267, opened 2026-09-19
fix(db): set synchronous=NORMAL on music_database connections

jcosmao:
## Summary

On a ~35,000-track library, the app went fully unresponsive roughly every
7 minutes of uptime: every endpoint, including `/status`, timed out. No
exception, no traceback in the logs: it wasn't crashing, it was starved.

`cpu.pressure` stayed flat at 0 the entire time, while `io.pressure some
avg10` sat around 29, a pure I/O stall, not a deadlock or an infinite loop.

## Root cause

`MusicDatabase._get_connection()` opens a new SQLite connection per call
and sets `journal_mode = WAL`, but never sets `synchronous`. WAL's default
is `synchronous=FULL`, which fsyncs on every commit, and unlike
`journal_mode`, `synchronous` is **not** persistent in the file, so it has
to be set per-connection.

Measured on the affected array, same file, same moment:

```
WAL + synchronous=FULL     152.7 ms per commit
WAL + synchronous=NORMAL     0.1 ms per commit
```

With a connection-per-call design, the library page's per-row work, an
import, and any repair job scan each open+commit repeatedly and serialize
behind that fsync. On a large library this queues into tens of seconds.
Gunicorn's `gthread` worker has a fixed thread pool: once enough request
threads are blocked waiting on that fsync, even a request that touches no
DB at all (like `/status`) can't get a free thread to run on. The liveness
probe then kills the pod as unresponsive, which just restarts the same
problem a few minutes later.

## Fix

Set `PRAGMA synchronous = NORMAL` on every connection in
`_get_connection()`, next to the existing `busy_timeout`/`foreign_keys`
pragmas. This is SQLite's documented-safe setting for WAL mode: an
OS-level crash or power loss can lose at most the last few
not-yet-checkpointed commits, never corrupt the database.

## Environment

- SoulSync 3.4.3 (also present in 3.4.0, unrelated to this specific upgrade)
- ~35,000 tracks, ~4,700 artists
- Self-hosted, Docker/Kubernetes, config + library on the same disk array as
  downloads/imports
- gunicorn, `gthread` worker, 8 threads, 1 worker process


Nezreka:
Thanks for digging into this. Setting it per connection makes sense. One thing before I merge: _ensure_wal_mode() logs and continues if WAL setup fails, so this would also set NORMAL when the DB is still in rollback mode. The WAL safety guarantee wouldn’t apply there.

Can you check the effective journal mode before setting NORMAL, and keep FULL otherwise? A small regression test covering both paths would be good too.
Also, “last few commits” isn’t a guaranteed limit. I’d change that to “recent commits may be lost after a power loss or OS crash.” Otherwise this looks good to me.

